PR #18, opened 2017-11-28
remove requests page

mirjamverloop:
asit is no longer in use


mirjamverloop:
@kgish I removed the "requests" as it is no longer in use. Can you please double check if everything is still fine?

kgish:
Deployed to test.

PR #161, opened 2018-07-17
bug/MET-1163-send empty map

mirjamverloop:


src/app/dataset/workflow/workflow.component.ts
```diff
@@ -353,9 +353,13 @@ export class WorkflowComponent implements OnInit {
     // mediaservice
     if (this.workflowForm.value['pluginMEDIA_PROCESS'] === true) {
       let connectionsMediaProcess = {};
+      console.log('limitConnectionsMEDIA_PROCESS', this.workflowForm.value['limitConnectionsMEDIA_PROCESS'].length);
       for (let c = 0; c < this.workflowForm.value['limitConnectionsMEDIA_PROCESS'].length; c++) {
```
stzanakis: This for loop and the one below for link checking seem identical. Maybe create a function and call that passing the array that needs to be "cleaned" from null/empty values?

mirjamverloop: Good suggestion, thanks!

PR #286, opened 2019-01-14
MET-1644 Progress Bar Improvements

andyjmaclean:


src/app/dashboard/executions/executiontable/executiontable.component.html
```diff
@@ -39,20 +39,13 @@
   <div class="progress">
     <div class="progress-status">
       <span class="progress-percentage"
-        >{{ 'step' | translate }} {{ execution.currentPluginIndex }}/{{
+        >{{ 'step' | translate }} {{ execution.currentPluginIndex + 1 }}/{{
```
heijmans: `npm run template-check` fails on this line:

    ERROR in src/app/dashboard/executions/executiontable/executiontable.component.html(42,10): : Object is possibly 'undefined'.

`execution.currentPluginIndex` might be undefined, so you should do something like `(execution.currentPluginIndex || 0) + 1` or `execution.currentPluginIndex ? execution.currentPluginIndex + 1 : 1`.

src/app/_helpers/calcprogress.ts
```diff
@@ -0,0 +1,13 @@
+import { WorkflowExecution } from '../_models';
+
+export function calcProgress(ongoing: WorkflowExecution): number {
+  if (!ongoing.currentPluginIndex) {
+    return 0;
+  }
+
+  const pct: number = (ongoing.currentPluginIndex / ongoing.metisPlugins.length) * 100;
```
heijmans: you can leave out the `: number` type annotations. but it's a matter of style, this is also fine.

src/app/_helpers/calcprogress.ts
```diff
@@ -0,0 +1,13 @@
+import { WorkflowExecution } from '../_models';
+
+export function calcProgress(ongoing: WorkflowExecution): number {
+  if (!ongoing.currentPluginIndex) {
```
heijmans: if the currentPluginx === 0, this condition will be triggered and 0 will be returned. But that is not what you want, you still want to calculate the subPct and return that. So you should do something like `if (ongoing.currentPluginIndex !== undefined)`.

PR #293, opened 2019-01-18
MET-1653 Widened XML / removed duplicate css

andyjmaclean:


src/assets/sass/scss/objects/_buttons.scss
```diff
@@ -200,10 +207,37 @@
...
+
+    &:active,
+    &:focus {
+      background: $eu-jade;
+      box-shadow: none;
+      color: $white;
+    }
+  }
+
+  &.btns-set-height-3em {
```
heijmans: i would prefer to use padding instead of an explicit height. that would work better if the button height would change or the content would wrap. the same comment applies to btns-set-height-5em. futhermore it is a bit strange that the class is called "3em" but the height is "3rem". that is confusing. i would prefer a more descriptive name, but that might be difficult to come up with. perhaps "btns-set-medium-padding" / "btns-set-small-padding".

src/assets/sass/pandora/base/_helpers.scss
```diff
@@ -11,12 +11,17 @@ body {
...
-  padding-left: 1.125rem;
-  padding-right: 1.125rem;
   width: 100%;
 
+  &.btns-set-right {
+    justify-content: flex-end;
+  }
+
   .btn-padding {
+    align-items: start;
```
heijmans: should be `align-items: flex-start;`

andyjmaclean: Made the change

src/assets/sass/scss/objects/_buttons.scss
andyjmaclean: Removed those classes

PR #298, opened 2019-01-25
Added helper unit tests

andyjmaclean:


src/app/_helpers/calcprogress.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+import { calcProgress } from './calcprogress';
+
+function makeWorkflowExecution(currIndex: number, progressStatuses: (number)[]): WorkflowExecution {
+  // tslint:disable:no-any
+  const fakePlugins: (any)[] = [];
+
+  progressStatuses.forEach((status) => {
+    fakePlugins.push({ executionProgress: { progressPercentage: status } });
+  });
+
```
heijmans: this is nicer with `map`:

    const fakePlugins = progressStatuses.map((status) => (
      { executionProgress: { progressPercentage: status } } as any
    ));

src/app/_helpers/calcprogress.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { WorkflowExecution } from '../_models';
+
+import { calcProgress } from './calcprogress';
+
+function makeWorkflowExecution(currIndex: number, progressStatuses: (number)[]): WorkflowExecution {
```
heijmans: you can just use `number[]` instead of `(number)[]`

src/app/_helpers/calcprogress.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  // tslint:disable:no-any
+  const fakePlugins: (any)[] = [];
```
heijmans: just use `any[]`

src/app/_helpers/statusclassfromplugin.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  // tslint:disable: no-any
+  return ({
+    pluginStatus: status,
+    executionProgress: { errors: 0 },
+  } as any) as PluginExecution;
+}
+
+function addErrorsToPlugin(pe: PluginExecution, errors: number): PluginExecution {
+  pe.executionProgress.errors = errors;
+  return pe;
```
heijmans: you should either return a new PluginExecution with the errors or you should modify the existing PluginExecution in place. For test code it's fine to just modify it in place and not return anything. it will also make the tests a little shorter.

heijmans:
you also need to run `npm run fix` to fix the tests.

src/app/_helpers/statusclassfromplugin.spec.ts
andyjmaclean: Done

src/app/_helpers/calcprogress.spec.ts
andyjmaclean: Done

src/app/_helpers/calcprogress.spec.ts
andyjmaclean: Done

src/app/_helpers/calcprogress.spec.ts
andyjmaclean: Done

PR #300, opened 2019-01-25
MET-1679 Removed trailing dot from report link href attributes

andyjmaclean:


src/app/shared/text-with-links/text-with-links.component.ts
```diff
@@ -30,4 +30,8 @@ export class TextWithLinksComponent {
     }
     this.parts = parts;
   }
+
+  public normaliseHref(href: string): string {
+    return href.replace(/[\.]$/, '');
```
heijmans: I think we should also drop comma's. So `/[\.,]$/`

You can also add a unit test for that.

src/app/shared/text-with-links/text-with-links.component.spec.ts
```diff
@@ -26,6 +26,12 @@ describe('TextWithLinksComponent', () => {
     expect(component.parts).toEqual(parts);
   }
 
+  it('removes trailing dots from link urls', () => {
+    const href = 'http://abc.com'
+    const hrefWithDot = href + '.'
+    expect(component.normaliseHref(hrefWithDot)).toEqual(href);
+  });
```
heijmans: I would also add a case where a URL without a dot is left unchanged.

PR #301, opened 2019-01-28
MET-1677 Added workflow parameters on hover

andyjmaclean:


src/app/dataset/workflow/workflow.component.scss
```diff
@@ -178,3 +178,15 @@
...
+.show-on-hover {
+  bottom: 0.2em;
+  display: none;
+  left: 1em;
+  position: relative;
+}
+
+.show-on-hover:hover,
+.plugin:hover ~ .show-on-hover {
+  display: inline-block;
```
heijmans: I would do this on hovering the `#import`.

src/app/dataset/workflow/workflow.component.ts
```diff
@@ -232,6 +232,18 @@ export class WorkflowComponent implements OnInit {
...
 
+  getImportSummary(): string {
+    let res: string;
+
+    if (this.harvestprotocol === 'HTTP_HARVEST') {
+      res = this.workflowForm.value!.url.trim();
+    } else {
+      res = this.workflowForm.value!.harvestUrl.trim();
+    }
+    res += ' ' + this.workflowForm.value!.setSpec;
```
heijmans: in case of a http harvest, the setspec is not used, so don't display it.

PR #303, opened 2019-02-04
MET-1690 Change cancel message (without using translation service)

andyjmaclean:


src/app/_services/workflow.service.ts
```diff
@@ -293,8 +294,9 @@ export class WorkflowService {
   }
 
   // show a prompt to cancel workflow
-  promptCancelThisWorkflow(id: string): void {
-    this.promptCancelWorkflow.emit(id);
+  promptCancelThisWorkflow(datasetId: string, datasetName: string): void {
+    const workflowDesc: WorkflowDesc = { id: datasetId, datasetName } as WorkflowDesc;
+    this.promptCancelWorkflow.emit(workflowDesc);
```
heijmans: you can just use: `this.promptCancelWorkflow.emit({ id: datasetId, datasetName });`. any object that has the right fields, will match `WorkflowDesc`. No need to cast.

src/app/_models/workflow-execution.ts
```diff
@@ -95,6 +95,11 @@ export interface WorkflowExecution {
   currentPluginIndex?: number;
 }
 
+export interface WorkflowDesc {
```
heijmans: i think CancelRequest is clearer. And add workflowExecutionId and datasetId.

src/app/_models/workflow-execution.ts
```diff
@@ -95,6 +95,12 @@ export interface WorkflowExecution {
   currentPluginIndex?: number;
 }
 
+export interface CancellationRequest {
+  id: string;
```
heijmans: perhaps change the id to workflowExecutionId?

heijmans:
check the failing checks...

PR #322, opened 2019-02-25
Feat/met 1771 use single report component

andyjmaclean:


src/app/dataset/lastexecution/lastexecution.component.ts
```diff
@@ -45,12 +46,8 @@ export class LastExecutionComponent {
...
   }
 
-  openReport(taskId: string, topology: TopologyName): void {
-    this.setReportRequest.emit({ taskId, topology });
-  }
-
-  openFailReport(errorMsg: string): void {
-    this.setReportMsg.emit(errorMsg);
+  openFailReport(topology?: TopologyName, taskId?: string, errorMsg?: string): void {
+    this.setReportMsg.emit({ topology, taskId, message: errorMsg });
```
mirjamverloop: I know that the reporting is subject to change, but you now have 3 files with the same function. Could you move it to a helper file/function, so in case you need to make changes, it is just in one place?

andyjmaclean: The line that's duplicated (or triplicated!) is a straight forwarding of parameters to an emitter.  If the signature of the emitter did have to change in the future then the helper function would have to change as well as all the calls to it, so in this case I'm not sure a helper is a good idea.

PR #323, opened 2019-02-25
MET-1769 Added sampling checkbox

andyjmaclean:


src/app/dataset/workflow/workflow.component.html
```diff
@@ -430,6 +429,60 @@
...
         /></label>
+        <div
+          class="form-group"
+          *ngIf="
+            !workflowForm.get('pluginLINK_CHECKING')!.disabled &&
+            workflowForm.get('pluginLINK_CHECKING')!.value
+          "
+        >
+          <div class="form-group-radios">
+            <div>
```
mirjamverloop: Is this div necessary?

src/app/dataset/workflow/workflow.component.html
```diff
@@ -430,6 +429,60 @@
...
+                  type="radio"
+                  formControlName="performSampling"
+                  value="true"
+                  [checked]="performSampling === true"
+                  (change)="changeLinkCheckSampling(true)"
+                />
+                {{ 'checksample' | translate }}
+              </label>
+            </div>
+            <div>
```
mirjamverloop: Is this div necessary?

src/app/dataset/workflow/workflow.component.html
andyjmaclean: This was copied and pasted from the harvest url radios.  I've removed the redundant div elements from both places.

src/app/dataset/workflow/workflow.component.html
andyjmaclean: This was copied and pasted from the harvest url radios.  I've removed the redundant div elements from both places.

PR #324, opened 2019-02-26
MET-1735 Use new tab opening datasets from dashboard

andyjmaclean:


src/app/dashboard/dashboardactions/dashboardactions.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="btns-set">
   <span class="btn-padding">
-    <a routerLink="/dataset/new" class="btn btn-left">
+    <a routerLink="/dataset/new" class="btn btn-left" target="_blank">
```
mirjamverloop: I am not sure if it is good practice to have a target and a routerlink.
See also: https://stackoverflow.com/a/42953823

mirjamverloop: And is this not causing any problems with the authentication/authorisation?

andyjmaclean: I see no problems.  From that same conversation:
  "Update2: As of v2.4.10 <a target="_blank" routerLink="/Page2"> works"

PR #328, opened 2019-03-08
MET-1802 Merge report buttons

andyjmaclean:


src/app/dataset/dataset.component.ts
```diff
@@ -94,20 +94,22 @@ export class DatasetComponent implements OnInit, OnDestroy {
   setReportMsg(req: SimpleReportRequest): void {
     if (req.message) {
       this.reportMsg = req.message;
-    } else if (req.taskId && req.topology) {
+    }
+    if (req.taskId && req.topology) {
```
mirjamverloop: I am not sure if I understand this correctly, but does this mean that the reportMsg (req.message) can be overwritten with "Report is empty"?

src/app/dataset/dataset.component.ts
```diff
@@ -94,20 +94,22 @@ export class DatasetComponent implements OnInit, OnDestroy {
...
       this.reportLoading = true;
-
       this.workflows.getReport(req.taskId, req.topology).subscribe(
         (report) => {
           if (report && report.errors && report.errors.length) {
             this.reportErrors = report.errors;
           } else {
             this.reportMsg = 'Report is empty.';
```
mirjamverloop: Move this to the translations perhaps?

src/app/dataset/dataset.component.ts
andyjmaclean: Instead of the "else" the report will show if there is either a message or a list of errors

PR #366, opened 2019-06-06
feat/MET-1963 Initial commit - workflow header

andyjmaclean:


src/app/dataset/workflow/workflow-header/workflow-header.component.ts
```diff
@@ -43,11 +44,31 @@ export class WorkflowHeaderComponent implements AfterViewInit {
 
   getAdjustableLabel(index: number): string {
     if (index === 0 && this.workflowForm) {
-      return this.workflowForm.value.pluginType;
+      if (this.workflowForm.value.pluginType) {
```
jochen-vermeulen: You can merge these two if statements and get rid of the duplicate return statement below.

andyjmaclean: Made the edit

andyjmaclean: (making it a net 2-line reduction!!!)

PR #381, opened 2019-07-15
MET-2063 Flag gaps in sequence

andyjmaclean:


src/app/_translate/lang-en.ts
```diff
@@ -62,6 +62,7 @@ export const LANG_EN_TRANS = {
   ftppassword: 'FTP/HTTP password',
   ftpuser: 'FTP/HTTP user',
   fullhistory: 'See full processing history',
+  gaperror: '406 Plugin Execution Not Allowed',
```
mirjamverloop: Is this message indeed correct? The story description mentions a message "This step is not allowed in this position. " as well.

src/app/dataset/workflow/workflow-header/workflow-header.component.scss
```diff
@@ -123,6 +123,9 @@ $border-colour-shortcuts: #dbd9d9;
       color: $white;
     }
   }
+  &.error {
+    border-color: red;
```
mirjamverloop: Can you use $eu-red here?

PR #390, opened 2019-07-30
MET-2130 Prevent workflow corruption

andyjmaclean:


src/app/dataset/workflow/workflow.component.ts
```diff
@@ -78,7 +78,9 @@ export class WorkflowComponent implements OnInit {
         .findIndex((plugin) => {
           return plugin.pluginType === 'LINK_CHECKING';
         });
-      if (index > -1) {
+      if (index === 0) {
+        this.rearrange(0, false);
+      } else if (index > -1) {
```
mirjamverloop: This looks a bit weird. As 0 is > larger than -1 as well. Does it work as you expected?

andyjmaclean: I've checked the else to be index > 0, which is logically the same but as you point out, cleaner.

PR #396, opened 2019-08-13
MET-2164 Labelling Fixes

andyjmaclean:


src/app/_translate/lang-en.ts
```diff
@@ -12,7 +12,7 @@ export const LANG_EN_TRANS = {
   cancelareyousure: 'Are you sure you want to cancel?',
   CANCELLED: 'Cancelled',
   cancelling: 'Cancelling',
-  'cancelled-by': 'Cancelled by user',
+  'cancelled-by': 'Cancelled by',
```
mirjamverloop: According to the task in Jira, it should stay "Cancelled by user", right?

PR #407, opened 2019-10-08
Feat/met 2138 test data server merged node v12

andyjmaclean:


src/app/dataset/actionbar/actionbar.component.spec.ts
```diff
@@ -77,8 +77,8 @@ describe('ActionbarComponent', () => {
     fixture.detectChanges();
     expect(workflows.promptCancelThisWorkflow).toHaveBeenCalledWith(
       '253453453',
-      undefined,
-      undefined
+      (undefined as unknown) as string,
+      (undefined as unknown) as string
```
mirjamverloop: This seems magic :) What does it do?

test-data/data/.env.example
```diff
@@ -0,0 +1,31 @@
...
+# The Google Tag Manager ID
+# Used by @nuxtjs/google-tag-manager to allow injection of scripts via google tag manager.
+GOOGLE_TAG_MANAGER_ID="GTM-XXXXXXX"
+
+# Redirect all record page requests to this URL path
+RECORD_PAGE_REDIRECT_PATH=/record-page-under-construction
+
+# If enabled, the content tier toggle will be displayed on the main search
+# results page. Disabled by default.
+ENABLE_CONTENT_TIER_TOGGLE=1
```
mirjamverloop: This seems like the Portal.js .env example file. How is it used in this project?

test-data/index.ts
```diff
@@ -0,0 +1,391 @@
...
+    ]
+  });
+}
+
+function routeToFile(response: ServerResponse, route: string): boolean {
+  response.setHeader('Content-Type', 'application/json;charset=UTF-8');
+
+  let regRes = route.match(/orchestrator\/proxies\/(\D+)\/task\/-?(\d+)\/report\/exists/);
+
+  if (regRes) {
```
mirjamverloop: I am not sure, but can this route.match - and the ones further down - go directly in the if? 

Eg: 
`if (route.match(/orchestrator\/proxies\/(\D+)\/task\/-?(\d+)\/report\/exists/)) { .... }`

README.md
```diff
@@ -6,8 +6,8 @@
 
 To run the Metis UI you need to provide a backend server to connect to. Copy `apisettings-test.ts` to `apisettings.ts` and fill in the URL's:
 
-    cp src/environments/apisettings-example.ts src/environments/apisettings.ts 
- 
+    cp src/environments/apisettings-example.ts src/environments/apisettings.ts
+
 If you are working at Europeana, just ask Andy MacLean or Mirjam Verloop for the right `apisettings.ts` file.
 
 Make sure you have `node` version 8, 10 or 11 and `npm` version 6.x:
```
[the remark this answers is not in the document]
andyjmaclean: Changed to 12

cypress/support/helpers.ts
```diff
@@ -1,59 +1,26 @@
...
-      'currentUser',
-      JSON.stringify({
-        user,
-        email: user.email,
-        token: user.metisUserAccessToken.accessToken
-      })
+    cy.request('POST', Cypress.env('dataServer') + '/users/data/authenticate.json').then(
+      (response) => {
+        const user = response.body as User;
+        w.localStorage.setItem(
```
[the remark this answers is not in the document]
andyjmaclean: It's been that way since the beginning, but I've added a cleanupUser function with an "afterEach" hook (note: "afterAll" wasn't found, but it's set up in a beforeEach so it's fine to remove it there)

src/app/dataset/actionbar/actionbar.component.spec.ts
andyjmaclean: This was to satisy the compiler with its latest libraries, as restoring it to return "undefined" (x2) results in the error:

ERROR in src/app/dataset/actionbar/actionbar.component.spec.ts(80,7): error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'string | AsymmetricMatcher<any>'.

test-data/index.ts
andyjmaclean: I have to access the match in the return of the function.  It can be accessed (but not declared) withing the if, so something like this is possible:

let x;
if(x=1 > 0){ console.log('yes...'); }

but is this desirable?  It's discouraged, allegedly, within Java...

mirjamverloop: Thanks, that makes sense, no need to change.

PR #416, opened 2019-10-25
Feat/met 2192 sonar qube fixes 4

andyjmaclean:


src/app/dashboard/filter-ops/filter-ops.component.ts
```diff
@@ -90,11 +90,9 @@ export class FilterOpsComponent implements FilterExecutionProvider {
 
   restoreGroup(group: string, callerIndex: number): void {
     this.getInputGroup(group).forEach((item) => {
-      if (item.index !== callerIndex) {
-        if (!item.valueIsSet()) {
-          if (item.getVal().length > 0) {
-            item.addParam();
-          }
+      if (item.index !== callerIndex && !item.valueIsSet()) {
+        if (item.getVal().length > 0) {
```
jochen-vermeulen: I think this may still generate a SonarQube warning because you could merge these two if statements.

src/app/dataset/workflow/workflow.component.ts
```diff
@@ -342,54 +341,75 @@ export class WorkflowComponent implements OnInit {
...
+  /* set all values in the FormGroup referenced by the conf fields to false
+  */
+  clearForm(): void {
     this.fieldConf.forEach((field) => {
       this.workflowForm.get(field.name)!.setValue(false);
     });
+  }
 
+  /** extractWorkflowParamsAlways
+  /* extract data values to the FormGroup if the plugin is not enabled
```
jochen-vermeulen: Not just if the plugin is not enabled, I think?

src/app/dashboard/filter-ops/filter-ops.component.ts
andyjmaclean: Done

PR #426, opened 2019-12-12
MET-1881 Flag max result reached

andyjmaclean:


andyjmaclean:
I spent time trying to parameterise the message (the "200" figure) but while this should be possible there's something in our setup preventing it from working.  I propose a separate ticket for translation parameterisation: the "Welcome, Valentine, let's ingest..." is hard-coded in the template and cannot be moved to the .en file until parameterisation is made possible

src/app/_translate/lang-en.ts
```diff
@@ -88,6 +88,7 @@ export const LANG_EN_TRANS = {
   loadmore: 'Load more',
   log: 'Log',
   mapping: 'Mapping',
+  'max-results-reached': "You've reached the limit of 200 displayed executions",
```
jochen-vermeulen: Perhaps, until we have implemented the new ticket regarding parametrized translation and so on, we could just remove all references to the number here? So perhaps:
"You've reached the limit of the number of executions that can be displayed."
The idea is that we are still going to experiment with the precise amount, and I don't think we should have something hard-coded.

src/app/dashboard/executionsgrid/executionsgrid.component.ts
```diff
@@ -119,11 +120,13 @@ export class ExecutionsgridComponent implements AfterViewInit, OnDestroy {
...
 
     this.overviewSubscription = polledOverviewData.subscribe(
-      ({ results, more }) => {
+      ({ results, more, maxResultCountReached }) => {
         this.hasMore = more;
         this.dsOverview = results;
         this.isLoading = false;
         this.isLoadingMore = false;
+        this.maxResultsReached = !!maxResultCountReached;
+        console.error('maxResultCountReached = ' + maxResultCountReached);
```
jochen-vermeulen: I think this line is your debugging line that should disappear?

src/app/_translate/lang-en.ts
andyjmaclean: Done

src/app/dashboard/executionsgrid/executionsgrid.component.ts
andyjmaclean: Hmm... I keep doing that.  Fixed.

PR #428, opened 2020-02-05
MET-2456 Locale key renaming / reorganisation

andyjmaclean:


src/app/_translate/lang-en.ts
```diff
@@ -4,226 +4,201 @@
...
-  'add-dataset-id': 'Add dataset id',
+export const LANG_EN_STATUS_TYPES = {
+  CANCELLED: 'Cancelled',
+  FAILED: 'Failed',
+  FINISHED: 'Finished',
+  INQUEUE: 'In queue',
+  RUNNING: 'Running'
+};
+
+export const LANG_EN_TRANS = Object.assign(LANG_EN_STATUS_TYPES, {
```
jochen-vermeulen: This copies all the constants in LANG_EN_STATUS_TYPES, which is not what we want and not what users of this structure would be expecting.

jochen-vermeulen: And your last commit solved it.

PR #441, opened 2020-05-29
Feat/met 2722 depublication

andyjmaclean:
Added the Depublication tab containing:
- grid (polled data)
- menu
- working modal dialogs
- data server upload / cache facility


src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  /** setPublicationFile
+  /*  post publication information
+  /*  @param {string} datasetId - the dataset id
+  /*  @param {File} file - file of record urls
+  */
+  setPublicationFile(datasetId: string, file: File): Observable<boolean> {
+    const url = `${apiSettings.apiHostCore}/records/${datasetId}`;
```
jochen-vermeulen, line 36: Can we rename `.../records/...` to `.../depublished_records/...`? Perhaps one day we wish to add other records for some reason.

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    const formData = new FormData();
+
+    formData.append('depublicationFile', file);
+
+    return this.http
+      .post(url, formData, {
+        headers: {
+          'Content-Type': file.type
```
jochen-vermeulen, line 44: I think this should be `multipart/form-data`.

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /** setPublicationInfo
+  /*  post publication information
+  /*  @param {string} datasetId - the dataset id
+  /*  @param {string} toDepublish - depublication record urls
+  */
+  setPublicationInfo(datasetId: string, toDepublish: string): Observable<boolean> {
+    const url = `${apiSettings.apiHostCore}/records/${datasetId}`;
+    return this.http.post<boolean>(url, { toDepublish }).pipe(this.errors.handleRetry());
```
jochen-vermeulen, line 64: I think in the backend I need the mime type to be set to distinguish it from the request above. How about `text/plain`?

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,106 @@
...
+  /*  @param {number} page - the pagination target
+  /*  @param {SortParameter} sort - optional - sort parameter
+  */
+  getPublicationInfo(
+    datasetId: string,
+    page: number,
+    sort?: SortParameter
+  ): Observable<Results<RecordPublicationInfo>> {
+    const sortParam = this.parseSortParameter(sort);
+    const url = `${apiSettings.apiHostCore}/records/${datasetId}?nextPage=${page}${sortParam}`;
```
jochen-vermeulen, line 87: 1. Why call it `nextPage` and not `page`?
2. What should we do with the `UNSET` option that `SortDirection` can have? I think you (the UI) should tell me what to sort on, not what not to sort on.

jochen-vermeulen: Oh, and:
3. Can we name the three sortable fields as follows:
- `recordId` (rather than `recordUrl` - it's not necessarily a URL)
- `depublicationStatus` (rather than `publicationStatus` - depublication is what we're after)
- `depublicationDate` (that one is fine as it is)

andyjmaclean: >     1. Why call it `nextPage` and not `page`?
> 
>     2. What should we do with the `UNSET` option that `SortDirection` can have? I think you (the UI) should tell me what to sort on, not what not to sort on.

The value UNSET never gets sent to the server.  It's included as states/classes are derived from the ENUM on individual headers.  When the header-group handler sees UNSET it nullifies the param information sent to the server.

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,106 @@
...
+import { map } from 'rxjs/operators';
+import { apiSettings } from '../../environments/apisettings';
+import { MoreResults, RecordPublicationInfo, Results, SortParameter } from '../_models';
+
+import { collectResultsUptoPage } from './service-utils';
+import { ErrorService } from './error.service';
+
+@Injectable({ providedIn: 'root' })
+export class DepublicationService {
+  constructor(private readonly http: HttpClient, private readonly errors: ErrorService) {}
```
jochen-vermeulen, line 13: Oh, and one thing... I think the httpClient takes care of authentication, right? The requests in this class will need the same authentication token in the header as the other requests.

andyjmaclean: I think it will work as is - based on what I see working for the dataset form

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -0,0 +1,257 @@
...
+  sortHeaderGroupConf = {
+    cssClass: 'grid-header',
+    items: [
+      {
+        translateKey: 'depublicationColUrl',
+        fieldName: 'recordId'
+      },
+      {
+        translateKey: 'depublicationColStatus',
+        fieldName: 'publicationStatus'
```
jochen-vermeulen, line 48: I think this should be `depublicationStatus`?

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    return p ? `&sortField=${p.field}&sortDirection=${p.direction.toLowerCase()}` : '';
+  }
+
+  /** getPublicationInfo
+  /*  retrieve publication information
+  /*  @param {string} datasetId - the dataset id
+  /*  @param {number} page - the pagination target
+  /*  @param {SortParameter} sort - optional - sort parameter
+  */
+  getPublicationInfo(
```
jochen-vermeulen, line 91: I added the option of searching as well to this service. The query parameter is called `searchQuery`.

src/app/_models/depublication.ts
```diff
@@ -0,0 +1,33 @@
...
+  DESC = 'DESC'
+}
+
+export class SortParameter {
+  field: string;
+  direction: SortDirection;
+}
+
+export enum DepublicationStatus {
+  PUBLISHED = 'PUBLISHED',
```
jochen-vermeulen, line 13: I think this should be called `PENDING`. Just to avoid all sorts of confusions. Or `NOT_DEPUBLISHED` if you prefer. First of all, we don't know whether the record is published (it may not even exist). Secondly, the aim is to depublish, not to publish.

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Component({
+  templateUrl: './sortable-header.component.html',
+  selector: 'app-sortable-header',
+  styleUrls: ['./sortable-header.component.scss']
+})
+export class SortableHeaderComponent {
+  isLocked = false;
+  statuses = Object.values(SortDirection);
+  classes = this.statuses.map((status: string) => status.toLowerCase());
```
jochen-vermeulen, line 12: Would it be better to have, in each enum member, a constant indicating the css class value rather than compute it from the enum value? That way you're not tied by that value and can give the class a more meaningful name.

src/app/_models/depublication.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface SortHeaderConf {
+  fieldName: string;
+  translateKey: string;
+  cssClass?: string;
+}
+
+export interface SortHeaderGroupConf {
+  cssClass: string;
+  items: Array<SortHeaderConf>;
+}
```
jochen-vermeulen, line 33: Same as above. I assume that the idea is to make this generic eventually?

src/app/dataset/depublication/sortable-group/sortable-group.component.html
```diff
@@ -0,0 +1,12 @@
+<ng-template #sortableGroupTemplate>
+  <app-sortable-header
+    (onSet)="onSetHandler($event)"
+    [conf]="{
+      translateKey: item.translateKey,
+      fieldName: item.fieldName,
+      cssClass: conf.cssClass
+    }"
+    *ngFor="let item of conf.items"
```
jochen-vermeulen, line 9: Would it be a good idea to rename the two different 'conf' variables that are referenced here so that their names are different?

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class DepublicationService {
+  constructor(private readonly http: HttpClient, private readonly errors: ErrorService) {}
+
+  /** handleUploadEvents
+  /*  record upload events
+  /*  @param {HttpEvent} event - event from the server
+  */
+  handleUploadEvents(event: HttpEvent<Object>): boolean {
```
jochen-vermeulen, line 19: According to my IDE, this can be a static function. Is that generally recommended?
Note: same for the functions `parseFilterParameter` and `parseSortParameter` below.

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,129 @@
...
+  parseFilterParameter(p?: string): string {
+    return p ? `&searchQuery=${p}` : '';
+  }
+
+  /** parseSortParameter
+  /*  return a url parameter representation of a SortParameter or an empty string
+  /*  @param {SortParameter} p - optional
+  */
+  parseSortParameter(p?: SortParameter): string {
+    return p ? `&sortField=${p.field}&sortDirection=${p.direction.toLowerCase()}` : '';
```
jochen-vermeulen, line 90: According to my IDE, it should be `p.direction.toString().toLowerCase()`. Is that needed/better? Or is this overkill?

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,129 @@
...
+  */
+  getPublicationInfo(
+    datasetId: string,
+    page: number,
+    sort?: SortParameter,
+    filter?: string
+  ): Observable<Results<RecordPublicationInfo>> {
+    const sortParam = this.parseSortParameter(sort);
+    const filterParam = this.parseFilterParameter(filter);
+    const url = `${apiSettings.apiHostCore}/depublished_records/${datasetId}?page=${page}${sortParam}${filterParam}`;
```
jochen-vermeulen, line 108: Would this not run the risk of invalid urls? Should the filter string not be escaped for URL parameters?
Perhaps there is some sort of URL builder we can use so that we don't have to actually manually create a query string?

src/app/_services/depublication.service.ts
```diff
@@ -0,0 +1,129 @@
...
+  /*  @param {SortParameter} sort - optional - sort parameter
+  /*  @param {string} filter - optional - sort parameter
+  */
+  getPublicationInfoUptoPage(
+    datasetId: string,
+    endPage: number,
+    sort?: SortParameter,
+    filter?: string
+  ): Observable<MoreResults<RecordPublicationInfo>> {
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
jochen-vermeulen, line 125: I assume that this comment disables that lint warning for the next line? In java we have the habit of putting another comment explaining why we think that a particular exclusion of the code quality rule is warranted here. Would that be a good idea here too?

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
```diff
@@ -0,0 +1,36 @@
...
+      field: this.conf.fieldName,
+      direction: this.statuses[this.current]
+    });
+    this.isLocked = false;
+  }
+
+  reset(): void {
+    if (!this.isLocked) {
+      this.current = 0;
+    }
```
jochen-vermeulen, line 34: Instead of checking isLocked, can you not pass the SortHeaderConf and check it against this.conf to know whether it is this header or another one that triggered the event? Then you don't need isLocked at all. 
Or perhaps your solution is less hackish?

src/app/dataset/depublication/sortable-header/index.ts
```diff
@@ -0,0 +1 @@
+export * from './sortable-header.component';
```
jochen-vermeulen, line 1: Do you foresee this ever being reused? Perhaps then sortable-group and sortable-header should be moved out of the depublication directory and somewhere shared? Like you're doing for the file upload component.

src/app/dataset/depublication/depublication-row/depublication-row.component.html
```diff
@@ -0,0 +1,13 @@
+<ng-template #depublicationTemplate>
+  <ng-template #empty_date><span class="date-and-time"></span></ng-template>
+  <span class="record-url">
+    <span class="">{{ record.recordId }}</span>
+  </span>
+  <span class="status">
+    <span class="">{{ record.depublicationStatus }}</span>
```
jochen-vermeulen, line 7: Should there be a translation operation here?

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -0,0 +1,275 @@
...
+    // stream for apply-filter
+    this.pollingRefresh = new Subject();
+    this.pollingRefresh.subscribe(() => {
+      pushPollDepublication += 1;
+    });
+
+    const loadTriggerDepublication = new BehaviorSubject(true);
+
+    const polledDepublicationData = loadTriggerDepublication.pipe(
+      merge(this.pollingRefresh), // user events comes into the stream here
```
jochen-vermeulen, line 242: My IDE says that this merge function is deprecated and should be used in a static way instead.

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -0,0 +1,275 @@
...
+  /*  returns an error if the form control value is the incorrect extension
+  /*  @param {FormControl} control - the input control to validate
+  */
+  validateFileExtension(control: FormControl): { [key: string]: boolean } | null {
+    const splitVal = (control.value ? control.value.name : '').split('.');
+    if (splitVal.length > 1 && splitVal[1].toLowerCase() !== 'txt') {
+      return { extension: true };
+    }
+    return null;
+  }
```
jochen-vermeulen, line 122: Both these functions can be made static according to my IDE.

src/app/dataset/depublication/depublication.component.html
```diff
@@ -0,0 +1,113 @@
...
+      'dialog-input': dialogInputOpen
+    }"
+  >
+    <div class="head">
+      {{ 'depublicationModalTitle' | translate }} <a class="btn-close" (click)="closeDialogs()"></a>
+    </div>
+    <ng-container *ngIf="dialogFileOpen">
+      <form [formGroup]="formFile" (ngSubmit)="onSubmitFormFile()">
+        <app-file-upload
+          formControlName="depublicationFile"
```
jochen-vermeulen, line 41: My IDE complains that the declaration of `formControlName` is not necessary (not used? not allowed?)

src/app/_models/depublication.ts
```diff
@@ -0,0 +1,33 @@
+export enum SortDirection {
+  UNSET = 'UNSET',
+  ASC = 'ASC',
+  DESC = 'DESC'
+}
+
+export class SortParameter {
+  field: string;
+  direction: SortDirection;
+}
```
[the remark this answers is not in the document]
andyjmaclean: Arguably, by the same reasoning, the sortable-header could be moved to the shared directory.  I considered that, but then thought it better to wait and see if there is to be any re-use first, and that there's benefit to having files / the contents of files grouped if there isn't.

I don't know what's "correct" but those were my thoughts.

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
andyjmaclean: I can see your point, but doesn't it clutter the code, and are the derived class names not going to be clear (by definition) on account of their one-to-one mapping with the status values?  If not would a prefix (or suffix) not bring clarity?

src/app/_models/depublication.ts
andyjmaclean: Which the optional cssClass?  Exactly.

src/app/dataset/depublication/sortable-header/index.ts
andyjmaclean: I mention my thinking on this above: moving it to "shared" would imply that it's shared when it isn't really.  That said... the file upload component is treated separately...

src/app/dataset/depublication/depublication-row/depublication-row.component.html
andyjmaclean: That would require keys corresponding to the ENUM values 'PENDING' and 'DEPUBLISHED' (and that would have the linter complain) or some mapping logic... just to arrive at the same visual outcome.

src/app/dataset/depublication/depublication.component.ts
andyjmaclean: That's probably true.  All the services are singleton classess (does your IDE say it for the others?) that could indeed be static, but the architecture uses the injection of an instance.

src/app/dataset/depublication/depublication.component.ts
andyjmaclean: See comment above

src/app/dataset/depublication/depublication.component.html
andyjmaclean: Providing `formControlName` exposes the required `onChange` hook needed for the handler.

src/app/dataset/depublication/depublication.component.ts
jochen-vermeulen: I think my IDE is referring to the fact that merge is no longer pipeable (if I understand correctly). See https://rxjs-dev.firebaseapp.com/guide/v6/migration.

src/app/dataset/depublication/depublication.component.ts
jochen-vermeulen: In the general case: I understand as unit testing in Java has the same effect, preventing some functions from becoming static. Please ignore this comment.

src/app/dataset/depublication/depublication-row/depublication-row.component.html
jochen-vermeulen: OK, so you're just outputting PENDING and DEPUBLISHED to the screen? Then I guess it is fine. Perhaps later at some point this should change, but we can see then.

src/app/_models/depublication.ts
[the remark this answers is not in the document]
jochen-vermeulen: I don't know what's correct either. 
OK, perhaps then for the time being we can keep sorting a part of the depublicaiton component. But we'll have to remember to move it to a shared space when we do reuse it later.

src/app/dataset/depublication/sortable-header/index.ts
jochen-vermeulen: As I commented above, let's leave it like it is for now.

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
jochen-vermeulen: Perhaps just adding a prefix would be the way to go indeed?

src/app/_services/depublication.service.ts
jochen-vermeulen: I assume unit tests are an issue here too?

src/app/_services/depublication.service.ts
andyjmaclean: That's no longer relevant.

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
andyjmaclean: Your suggestion was my initial approach - I thought this way was clearer

src/app/_services/depublication.service.ts
andyjmaclean: Removed - it's been changed so it isn't needed

src/app/_services/depublication.service.ts
andyjmaclean: used `encodeURIComponent()`

src/app/dataset/depublication/sortable-header/sortable-header.component.ts
andyjmaclean: Added the prefix `sort-`

PR #443, opened 2020-06-16
validate input for format and parent dataset

andyjmaclean:


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -92,13 +92,43 @@ export class DepublicationComponent implements OnDestroy {
...
+  /*  returns an error if the form control value includes invalid record ids
+  /*  @param {FormControl} control - the input control to validate
+  */
+  validateRecordIds(control: FormControl): { [key: string]: boolean } | null {
+    const val = control.value || '';
+    let invalid = false;
+    let misref = false;
+
+    const reg = /(http(s)?:\/\/)?([a-zA-Z]*\/)?([0-9]*\/)?[a-zA-Z\d_]*$/;
+    const reg2 = new RegExp('\\b' + this._datasetId + '\\b');
```
jochen-vermeulen, line 115: And here we should ensure that the database id occurs at the right segment. How about capturing it as a group in the first regular expression and verifying it? Building on my previous comment, it would look like this:
`/^((http(s)?:\/\/)?([^\s\/]+\/)*(?<dataset_capture>[a-zA-Z\d]+)\/)?\w+$/`
And then we check the value of `dataset_capture` and see if it equals `this._datasetId` (if it even occurs).

src/app/dataset/depublication/depublication.component.spec.ts
```diff
@@ -117,15 +118,56 @@ describe('DepublicationComponent', () => {
...
+      ).toBeFalsy();
+      expect(
+        component.validateRecordIds(
+          frmCtrl(`
+        https://path/${datasetId}/${recordId}/
+        https://path/${datasetId}/${recordId}
+      `)
+        )
+      ).toBeTruthy();
+    });
```
jochen-vermeulen, line 169: I can come up with some others. How about checking the following:
- `http://www.server.com//path1/path2/${datasetId}/${recordId}` (should pass)
- `/${datasetId}/not_the_dataset/${recordId}` (should fail)
- values with spaces in the url (should fail)

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -109,24 +109,22 @@ export class DepublicationComponent implements OnDestroy {
...
-    let misref = false;
-
-    const reg = /(http(s)?:\/\/)?([a-zA-Z]*\/)?([0-9]*\/)?[a-zA-Z\d_]*$/;
-    const reg2 = new RegExp('\\b' + this._datasetId + '\\b');
-
+    const reg = new RegExp(
+      '(^(http(s)?://[^\\s]*)?(^(\\/){1})?([^\\s\\/]+\\/)*(' +
+        this.datasetId +
+        '\\/)+)?[^\\s\\/:\\.]+'
+    );
```
jochen-vermeulen, line 116: I found some obvious problems. I don't think we can aim for perfection here, but perhaps we can at least catch these fairly dramatic cases (assume datasetId = "3"):

'https://path/43/asd' (or any dataset ID ending with 3)
'https:/a/3/c'
'htps:/a/3/c'
'a-_$&#%@)}[*!@#'
'https:///////path//////3/asd'
'https:///path/3/asd'

I tweaked your regular expression a bit, and came up with this:

```
    var reg = new RegExp('^(((http(s)?:\\/\\/)|\\/)?([^\\s\\/:]+\\/)*(' +
        this.datasetId +
        '\\/)+)?[\\w]+$'
    );
```

This seems to prevent all these cases while still allowing the good ones I tried.

PR #444, opened 2020-06-17
Model changes / basic view changes

andyjmaclean:


src/app/dataset/generalinfo/generalinfo.component.html
```diff
@@ -1,6 +1,9 @@
 <div class="metis-dataset-info-block" *ngIf="datasetData">
   <div class="dataset-name">
     <span class="svg-icon-dataset"></span> {{ datasetData.datasetName }}
+    <span *ngIf="isCurrentlyDepublished" class="mark-depublished">{{
+      'depublished' | translate
+    }}</span>
```
jochen-vermeulen, line 6: Shouldn't this be a 3-way choice? 
1. Currently published
2. Currently depublished
3. Neither (has never been published).

src/app/_models/harvest-data.ts
```diff
@@ -4,9 +4,11 @@ export interface HarvestData {
   lastPreviewDate: string;
   lastPreviewRecords: number;
   lastPreviewRecordsReadyForViewing: boolean;
-
   firstPublishedDate: string;
+  isCurrentlyDepublished?: boolean;
+  lastDepublishedDate?: string;
   lastPublishedDate: string;
+  lastDepublishedRecords?: number;
```
jochen-vermeulen, line 11: I don't know whether these are the right fields. Have you agreed them with Simon?

src/app/dataset/generalinfo/generalinfo.component.html
andyjmaclean: I thought we said it was reducible to a boolean?

jochen-vermeulen: I thought we discussed a three way boolean (true, false or null)... but perhaps an enum is better.

andyjmaclean: The design I see indicates that it either shows "depublished" or it doesn't.  I've made the underlying model an enum but reduced this to a boolean for this section of the UI

src/app/_models/harvest-data.ts
andyjmaclean: As per our private conversation with Simon we're keeping the bottom 3 and changing the first to an enumeration.  I've renamed the variable to "currentPublicationStatus"

src/app/dataset/generalinfo/generalinfo.component.ts
```diff
@@ -11,7 +11,7 @@ import { Dataset, HarvestData } from '../../_models';
 export class GeneralinfoComponent {
   @Input() datasetData: Dataset;
 
-  isCurrentlyDepublished?: boolean;
+  isCurrentlyDepublished = false;
```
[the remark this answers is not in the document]
andyjmaclean: It's showing "published" or "depublished".  If you're using the dummy data server try switching between 
 * http://localhost:4280/dataset/depublication/0
and
 * http://localhost:4280/dataset/depublication/1

PR #445, opened 2020-06-19
Initial commit

andyjmaclean:


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -228,6 +245,15 @@ export class DepublicationComponent implements OnDestroy {
     }
   }
 
+  onDepublishDataset(): void {
+    this.depublications.depublishDataset(this._datasetId).subscribe((success: boolean) => {
+      if (success) {
+        this.depublicationComplete = true;
```
jochen-vermeulen, line 251: Maybe set it to false before triggering the depublication? Just in case you're doing it more than once... (which really shouldn't happen, but you never know)

src/app/dataset/depublication/depublication.component.html
```diff
@@ -9,6 +9,9 @@
     <li *ngIf="isSaving">
       <span class="svg-icon-spin showing"></span>
     </li>
+    <li>
+      <a (click)="onDepublishDataset()">{{ 'depublishDataset' | translate }}</a>
+    </li>
```
jochen-vermeulen, line 14: I think now it is just one button, right (not a drop-down)? Later we will add drop-down options to it (in a future ticket). Just so that you're aware.

src/app/dataset/depublication/depublication.component.ts
andyjmaclean: I'm using this.depublicationComplete to show the message in the grid as to how many records were depublished.  It's true that this will need cleared, but there's no obvious time to do it.  (perhaps when they leave the tab and return to it?)

src/app/dataset/depublication/depublication.component.html
andyjmaclean: Yep.  I know.

src/app/dataset/depublication/depublication.component.ts
jochen-vermeulen: Well, if the only result is that the message just stays until it is replaced by the next one, I guess it is fine. In reality this is really not very likely.

PR #447, opened 2020-06-30
Added checkboxes

andyjmaclean:


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -310,11 +355,14 @@ export class DepublicationComponent implements OnDestroy {
           }
         });
       })
-    ) as Observable<MoreResults<RecordPublicationInfo>>;
+    ) as Observable<MoreResults<RecordDepublicationInfo>>;
```
jochen-vermeulen, line 358: According to my IDE this assertion is unnecessary and it does not change the type of the expression.

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -310,11 +355,14 @@ export class DepublicationComponent implements OnDestroy {
...
 
     this.depublicationSubscription = polledDepublicationData.subscribe(
       ({ results, more }) => {
-        this.depublicationData = results;
+        this.depublicationData = results.map((entry: RecordDepublicationInfoDeletable) => {
+          entry.deletion = this.depublicationDeletions.indexOf(entry.recordId) > -1;
+          return entry as RecordDepublicationInfoDeletable;
```
jochen-vermeulen, line 364: According to my IDE also here the assertion is not necessary and the type of the expression is not changed.

src/app/dataset/depublication/sortable-header/sortable-header.component.html
```diff
@@ -1,3 +1,8 @@
-<span class="{{ conf.cssClass }} {{ classes[current] }}" (click)="valueBump()"
-  ><span class="state-arrow">{{ conf.translateKey | translate }}</span>
+<span class="{{ conf.cssClass }} {{ classes[current] }}" (click)="valueBump()">
+  <ng-template #empty
```
jochen-vermeulen, line 2: Maybe choose a different name than `empty`? Perhaps `selectAll` or something like that?

PR #449, opened 2020-07-01
Add default fitness option

andyjmaclean:


src/app/dataset/dataset.component.ts
```diff
@@ -366,16 +366,27 @@ export class DatasetComponent implements OnInit, OnDestroy {
...
+    }
+    let cssClass = '';
+    let warning = '';
     if (status === PublicationFitness.UNFIT) {
-      return this.translate.instant('datasetUnpublishableBanner');
+      warning = this.translate.instant('datasetUnpublishableBanner');
     } else if (status === PublicationFitness.PARTIALLY_FIT) {
-      return this.translate.instant('datasetPartiallyUnpublishableBanner');
+      warning = this.translate.instant('datasetPartiallyUnpublishableBanner');
+      cssClass = 'partial-fitness';
```
jochen-vermeulen, line 385: Perhaps more consistent with the rest of the code to use the translation option as a pipe in the html document?

src/app/dataset/dataset.component.html
```diff
@@ -36,10 +36,13 @@
       </div>
       <div class="metis-dashboard-right">
         <div
-          *ngIf="publicationFitnessWarning(datasetData.publicationFitness); let warning"
-          class="unfit-to-publish"
+          *ngIf="
+            publicationFitnessWarningAndClass(datasetData.publicationFitness);
+            let warningAndClass
+          "
+          class="unfit-to-publish {{ warningAndClass.cssClass }}"
```
jochen-vermeulen, line 43: Since the css class `partial-fitness` extends `unfit-to-publish` do you even need two classes here? Perhaps you can move `unfit-to-publish` to be the UNFIT alternative in the `warningAndClass` object?

andyjmaclean: Done

PR #451, opened 2020-07-08
Remove redundant data call / supply correct parameters for recovered …

andyjmaclean:


src/app/dataset/preview/preview.component.ts
```diff
@@ -98,7 +99,8 @@ export class PreviewComponent implements OnInit, OnDestroy {
     this.editorConfig = this.editorPrefs.getEditorConfig(true);
     this.nosample = this.translate.instant('noSample');
 
-    this.executionsFilterTimer = timer(0, environment.intervalStatusMedium).subscribe(() => {
+    this.serviceTimer = timer(0, environment.intervalStatusMedium);
+    this.executionsFilterSubscription = timer(0, environment.intervalStatusMedium).subscribe(() => {
```
jochen-vermeulen, line 103: Shouldn't this be `this.serviceTime.subscribe(...`?
Or do you need it to be two instances here?

andyjmaclean: You're quite right - I created the variable then forgot to use it.  Corrected now.

PR #452, opened 2020-07-13
Button / checkbox disabling

andyjmaclean:


src/app/dataset/depublication/depublication.component.html
```diff
@@ -137,12 +139,14 @@
...
   </span>
 
-  <ng-container *ngFor="let record of depublicationData">
-    <app-depublication-row
-      #dataWrapper
-      [record]="record"
-      (checkEvents)="processCheckEvent($event)"
-    ></app-depublication-row>
-    <ng-content *ngTemplateOutlet="dataWrapper.depublicationTemplate"></ng-content>
+  <ng-container *ngIf="!depublicationComplete">
```
jochen-vermeulen, line 142: It looks like the table content becomes invisible if a depublication finishes. Is that what you are doing here? I don't think that's right.

andyjmaclean: I've removed it.

PR #456, opened 2020-07-21
Refactor buttons to menu / add control to depubish all / label tweak …

andyjmaclean:
Refactor buttons to menu / add control to depubish all / label tweak / test tweak


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -323,24 +341,28 @@ export class DepublicationComponent extends DataPollingComponent implements OnDe
...
-        this.depublicationSelections = [];
-        this.pollingRefresh.next(true);
-        this.isSaving = false;
-      },
-      (err: HttpErrorResponse): void => {
-        this.onError(err);
-      }
-    );
+    this.depublications
+      .depublishRecordIds(this._datasetId, all ? [] : this.depublicationSelections)
```
jochen-vermeulen, line 349: I know it is not needed (because the button would be disabled) but perhaps we could do a sanity check here that if `all` is false but `this.depublicationSelections` is empty, it doesn't send an empty list to the server thus causing all records to be deleted by accident?

src/app/dataset/depublication/depublication.component.html
```diff
@@ -21,21 +21,35 @@
...
+              (click)="onDepublishRecordIds()"
+              [attr.disabled]="depublicationSelections.length === 0"
+              >{{
+                'depublishRecordsIds' +
+                  (depublicationSelections.length === 1 ? 'Single' : 'Plural') | translate
+              }}</a
+            >
+          </li>
+          <li>
+            <a (click)="onDepublishRecordIds()">{{ 'depublishRecordsIdsAll' | translate }}</a>
```
jochen-vermeulen, line 39: Should this not be `(click)="onDepublishRecordIds(true)"`? The boolean parameter seems to be missing?

jochen-vermeulen: Ah, one more thing: Perhaps this menu option (as well as the one above) should also have `depublishDatasetDisabled` as disabled condition (just like the entire dataset option). The same rules apply, after all.

src/app/dataset/depublication/depublication.component.ts
andyjmaclean: Added the check

src/app/dataset/depublication/depublication.component.html
andyjmaclean: I've removed the condition from the item and added it to the entire menu

PR #457, opened 2020-07-21
Inverse disabling logic

andyjmaclean:


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -118,7 +118,9 @@ export class DepublicationComponent extends DataPollingComponent implements OnDe
   */
   setSelection(val: boolean): void {
     this.depublicationRows.forEach((row) => {
-      row.onChange(val);
+      if (!val || (val && !row.checkboxDisabled())) {
```
jochen-vermeulen, line 121: I think this condition is logically the same as `(!val || !row.checkboxDisabled())`. What do you think?

PR #467, opened 2020-08-11
Handle 401 / remove redundant check

andyjmaclean:


src/app/_services/error.service.ts
```diff
@@ -22,7 +24,7 @@ export class ErrorService {
   /* @param {object} err - details of error
   */
   handleError(err: HttpErrorResponse): HttpErrorResponse | false {
-    if (err.status === 401 || err.error.errorMessage === 'Wrong access token') {
+    if (err.status === 400 || err.status === 401) {
```
jochen-vermeulen, line 27: As discussed, we will not check for the 400 errors.

PR #475, opened 2020-08-31
Enable throwError rule

andyjmaclean:


src/app/_helpers/cache.spec.ts
```diff
@@ -38,11 +38,11 @@ describe('single cache', () => {
...
   it('should not cache an error', () => {
-    const fn = jasmine.createSpy().and.callFake(() => throwError('wrong'));
+    const fn = jasmine.createSpy().and.callFake(() => throwError(new Error('wrong')));
     const cache = new SingleCache<number>(fn);
-    expect(gatherError(cache.get())).toEqual('wrong');
-    expect(gatherError(cache.get())).toEqual('wrong');
-    expect(gatherError(cache.get())).toEqual('wrong');
+    new Array(3).fill(null).map(() => {
+      expect(gatherError(cache.get())).not.toEqual('wrong');
+    });
```
jochen-vermeulen, line 45: My sonarlint complains about mapping as the result is not used. It suggests using `forEach`. 
Why not just do some easy-to-read for loop? Perhaps:
```
    new Array(3).forEach(() => {
      expect(gatherError(cache.get())).not.toEqual('wrong');
    });
```
Or even
```
[1,2,3].forEach(() => {
  expect(gatherError(cache.get())).not.toEqual('wrong');
}
```

src/app/_helpers/cache.spec.ts
```diff
@@ -100,13 +100,11 @@ describe('keyed cache', () => {
...
+    const fn = jasmine.createSpy().and.callFake(() => throwError(new Error('wrong')));
     const cache = new KeyedCache<string>(fn);
-    expect(gatherError(cache.get('1'))).toEqual('wrong');
-    expect(gatherError(cache.get('2'))).toEqual('wrong');
-    expect(gatherError(cache.get('3'))).toEqual('wrong');
-    expect(gatherError(cache.get('2'))).toEqual('wrong');
-    expect(gatherError(cache.get('3'))).toEqual('wrong');
+    new Array(5).fill(null).map((i) => {
+      expect(gatherError(cache.get(`${i + 1}`))).not.toEqual('wrong');
+    });
```
jochen-vermeulen, line 107: See above.

src/app/_services/error.service.ts
```diff
@@ -56,6 +56,7 @@ export class ErrorService {
           }),
           take(this.numberOfRetries)
         ),
+        // eslint-disable-next-line rxjs/throw-error
```
jochen-vermeulen, line 59: I know we discussed this, but only within the context of your mocked service. This is actual production code. 
Is there no way to do this more elegantly? Why not pass an instance of `HttpErrorResponse`?

src/app/_services/error.service.ts
```diff
@@ -60,7 +60,7 @@ export class ErrorService {
         throwError({
           status: 0,
           error: { errorMessage: this.translate.instant('errorRetryFailed') }
-        })
+        } as HttpErrorResponse)
```
jochen-vermeulen, line 63: The point is that you can now remove the disable lint rule statement.
Also: is this not a solution for the other places where you disabled this lint rule?

andyjmaclean: Of course!  Done.

PR #478, opened 2020-09-03
Refactor DatasetLog / address flakey unit test

andyjmaclean:


src/app/dataset/datasetlog/datasetlog.component.ts
```diff
@@ -1,29 +1,33 @@
...
-  ) {}
+  ) {
+    super();
+  }
 
   @Output() closed = new EventEmitter<void>();
 
   logMessages?: SubTaskInfo[];
   logPerStep = 100;
   logStep = 1;
```
jochen-vermeulen, line 30: My IDE says that this field is unused.

jochen-vermeulen (changes requested):

src/app/dataset/datasetlog/datasetlog.component.ts
andyjmaclean: Weird that mine does not.  Removed.

PR #481, opened 2020-09-14
Add Depublish label to renameWorkflowPipe

andyjmaclean:


src/app/_translate/renameworkflow.pipe.ts
```diff
@@ -8,6 +8,7 @@ import { Pipe, PipeTransform } from '@angular/core';
 const workflowNames: { [key: string]: string } = {
   DEPUBLISH_DATASET: 'Depublish Dataset',
   DEPUBLISH_RECORDS: 'Depublish Records',
```
jochen-vermeulen, line 10: Are these two then still needed?

andyjmaclean: Apparently not!

PR #486, opened 2020-09-17
Conditionally show fitness ops / Cypress test refactor

andyjmaclean:


andyjmaclean:
Automated checks will pass when merged with [this](https://github.com/europeana/metis-ui/pull/484)

src/app/dataset/datasetform/datasetform.component.html
```diff
@@ -177,7 +177,7 @@
         <label class="dataset-label" for="notes">{{ 'notes' | translate }}</label>
         <textarea id="notes" formControlName="notes" (blur)="saveTempData()"></textarea>
       </div>
-      <div class="form-group">
+      <div class="form-group" *ngIf="!isNew">
```
jochen-vermeulen, line 180: This is good, but the frontend still sends the default "not fit for publication" to the backend (see `datasetform.component.ts` line 242). Perhaps we can send another default (i.e. `PublicationFitness.FIT`)?

jochen-vermeulen (changes requested):

PR #491, opened 2020-09-24
Select all checkbox fix binding

andyjmaclean:


src/app/dataset/depublication/depublication.component.ts
```diff
@@ -134,6 +142,20 @@ export class DepublicationComponent extends DataPollingComponent {
         return deletionInfo.recordId !== recId;
       });
     }
+    this.checkAllAreSelected();
+  }
+
+  /** checkAllAreSelected
+  /*  sets allSelected variable according to check states
+  */
+  checkAllAreSelected(): void {
```
jochen-vermeulen, line 151: I found the following interesting flow.
1. remove all records from the list.
2. select the 'select all' checkbox.
3. add records from a file.
4. All new records are deselected, but the select all checkbox is still selected.
It's not very dramatic, but still...

src/app/dataset/depublication/depublication.component.ts
```diff
@@ -150,10 +152,12 @@ export class DepublicationComponent extends DataPollingComponent {
   */
   checkAllAreSelected(): void {
     if (this.depublicationRows) {
+      this.selectAllDisabled = this.depublicationRows.length === 0;
```
jochen-vermeulen, line 155: Wouldn't it be better to let this depend on the size of enabledRows instead of depublicationRows? Will give us some nicer behavior in case all rows are disabled.

PR #492, opened 2020-09-25
Alter copy / include i18n in dist task

andyjmaclean:


src/app/dataset/depublication/depublication.component.html
```diff
@@ -16,7 +16,7 @@
   <p
     i18n="confirm all record id depublish|show modal to get user confirmation@@confirmRecordAllIdsDepublish"
   >
-    Are you sure you want to cancel record(s) in "{{ datasetName }}"?
+    Are you sure you want to depublish record(s) in "{{ datasetName }}"?
```
jochen-vermeulen, line 19: I think you also need to reword the message above (line 11).

jochen-vermeulen (changes requested):

PR #493, opened 2020-09-28
Hide / remove drag ghost element

andyjmaclean:


src/app/dataset/workflow/workflow-header/workflow-header.component.ts
```diff
@@ -157,20 +158,25 @@ export class WorkflowHeaderComponent implements AfterViewInit {
...
     }
   }
 
   /** dragEnd
   /* reset drag-handling variables
   */
   dragEnd(): void {
+    if (this.ghostClone) {
+      this.ghostClone.remove();
+    }
```
jochen-vermeulen, line 179: Would it make sense to set this.ghostClone to null, just to be safe?

src/app/dataset/workflow/workflow-header/workflow-header.component.ts
```diff
@@ -157,20 +158,25 @@ export class WorkflowHeaderComponent implements AfterViewInit {
...
-      n.style.top = '-1000px';
+      n.style.left = '-10000px';
       n.style.transform = 'scale(1)';
       n.style.backgroundSize = 'contain';
       n.style.height = '40px';
+      n.style.position = 'relative';
       n.style.width = '40px';
       document.body.appendChild(n);
       e.dataTransfer.setDragImage(n, 20, 20);
+      this.ghostClone = n;
```
jochen-vermeulen, line 169: Instead of using temporary variable n, why not use this.ghostClone directly?

src/app/dataset/workflow/workflow-header/workflow-header.component.ts
andyjmaclean: I would have to redeclare it as optional, and that's unnecessary as the only time it's read it's already inside a null check

`ERROR in src/app/dataset/workflow/workflow-header/workflow-header.component.ts(179,7): error TS2322: Type 'null' is not assignable to type 'Element'.`

src/app/dataset/workflow/workflow-header/workflow-header.component.ts
andyjmaclean: Short answer: we'd need to upgrade typescript!

PR #496, opened 2020-10-01
Refactor isLoading logic / load comparison editor on prefill

andyjmaclean:


src/app/_models/workflow-execution.ts
```diff
@@ -107,9 +107,14 @@ export interface WorkflowExecutionHistoryList {
...
-export interface PreviewFilters {
+export interface PreviewFilter {
   executionId?: string;
   pluginType?: PluginType;
+}
+
+export interface PreviewFilters {
+  basic: PreviewFilter;
+  comparison?: PreviewFilter;
   startedDate?: string;
```
jochen-vermeulen, line 118: Perhaps some better naming? What about baseFilter, comparisonFilter and baseStartedDate?

src/app/dataset/dataset.component.ts
```diff
@@ -62,7 +62,7 @@ export class DatasetComponent extends DataPollingComponent implements OnInit {
 
   showPluginLog?: PluginExecution;
   tempXSLT?: string;
-  previewFilters: PreviewFilters = {};
+  previewFilters: PreviewFilters = { basic: {}, comparison: {} };
```
jochen-vermeulen, line 65: You could also leave comparison equal to null, right? It's optional.

src/app/dataset/preview/preview.component.ts
```diff
@@ -248,58 +264,78 @@ export class PreviewComponent extends SubscriptionManager implements OnInit, OnD
...
+          const prvCmp = this.previewFilters.comparison;
+
+          if (prefilling && prvCmp && prvCmp.pluginType && prvCmp.executionId) {
+            const comparisonPluginType = prvCmp.pluginType;
+            const comparisonExecutionId = prvCmp.executionId;
+            this.getXMLSamplesCompare(comparisonPluginType, comparisonExecutionId, prefilling);
+            this.isLoadingSamples = false;
+          } else {
+            this.isLoadingSamples = false;
+          }
```
jochen-vermeulen, line 304: First of all, the `isLoadingSamples` assignment can be extracted from both branches.

Second: I feel the code would be a lot nicer if the call to `getXMLSamplesCompare` is taken out of this method. It is only triggered by the call from `previewFilters()` (line 374 below) and it would mean this simplifies the code here enormously. Also, this way we don't have to do Subscription within Subscription. But the question is: is there a time dependency here?

andyjmaclean: On your first point I concur.  On your second, I chained the call to avoid a race condition.  In the course of variable tidying the execution flow may now allow that - I'll check....

PR #497, opened 2020-10-02
Use Modal-Confim dialog instead of custom code

andyjmaclean:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "metis-ui",
-  "version": "2.0-development",
+  "version": "1.7.0-development",
```
jochen-vermeulen, line 3: This doesn't quite make sense to me...

src/app/app.component.html
```diff
@@ -2,23 +2,13 @@
...
-        <button class="button" type="button" (click)="closePrompt()">{{ 'no' | translate }}</button>
-        <button class="button" type="button" (click)="cancelWorkflow()">
-          {{ 'yes' | translate }}
-        </button>
-      </p>
-    </div>
-  </div>
-</div>
+<app-modal-confirm id="{{ modalConfirmId }}" title="{{ 'cancel' | translate }}">
+  <ng-container *ngIf="cancellationRequest">
```
jochen-vermeulen, line 6: This shouldn't have a `*gnIf` condition, right? The modal itself (`modal-confirm-component.html`) has a condition and that should be sufficient? Just like the three modals in `depublication.component.html'.

package.json
andyjmaclean: I'd created a dev branch for development with that single change (and since deleted it) - this was to restore...

src/app/app.component.html
andyjmaclean: The display logic is handled by `ModalConfirmService.open`, which calls open on the `ModalCofirmComponent`, setting the boolean `show`.  This is picked up by the `ngIf` in `src/app/shared/modal-confirm/modal-confirm.component.html`

jochen-vermeulen: Exactly, so my point is that your `<ng-container *ngIf="cancellationRequest">` here is just not needed, right?

andyjmaclean: The if logic around cancellationRequest is for the display message which references it twice.

jochen-vermeulen: Ah, I see now.
What do you think? Just to be sure: it should never be null, right (if the modal is showing, I mean)? Should we take steps to avoid the (unexpected) situation in which the popup shows, but the text message does not? E.g. put a default message or check for null when compiling the message?
We can also just leave it like this. Whatever you think is best.

andyjmaclean: The displays are independently controlled by variables that are seemingly unconnected in the template, but we expect them to coincide because of this code in the component:
```
      this.workflows.promptCancelWorkflow.subscribe((cancellationRequest: CancellationRequest) => {
        this.cancellationRequest = cancellationRequest;
        if (cancellationRequest.workflowExecutionId) {
          this.showWrapper = true;
        }
      })
```
This is the only place that `showWrapper` is set to true, so the unexpected situation isn't one that can (currently) happen.  In the future that may change (not sure how, but it may) but with nothing concrete to future-proof against I'd be against adding extra logic.

jochen-vermeulen: OK. Let's keep it this way then. I'll test and merge tomorrow.

PR #501, opened 2020-10-08
Feat/met 2976 upgrade angluar

andyjmaclean:


src/app/dataset/reportsimple/reportsimple.component.ts
```diff
@@ -82,12 +82,18 @@ export class ReportSimpleComponent {
...
+    const doIfSelectionPresent = (fn: (sel: Selection) => void): void => {
+      const selection = window.getSelection();
+      if (selection) {
+        fn(selection);
+      }
+    };
+    doIfSelectionPresent((sel: Selection) => sel.removeAllRanges());
     const range = document.createRange();
     range.selectNode(element);
-    window.getSelection().addRange(range);
```
jochen-vermeulen, line 88: Isn't this the same as `window.getSelection()?.addRange(range);`?
Same for two lines down (line 96).

src/app/shared/subscription-manager/subscription.manager.ts
```diff
@@ -1,9 +1,10 @@
 /** SubscriptionManager
 /*  stores references to subscriptions / unsubscribes from them on destroy
 */
-import { OnDestroy } from '@angular/core';
+import { Directive, OnDestroy } from '@angular/core';
 import { Subscription } from 'rxjs';
 
+@Directive()
```
jochen-vermeulen, line 7: Doesn't a directive need a selector so that it can be referenced in the HTML/DOM?

andyjmaclean: This is a superclass and only child classes are referenced by the DOM, so it would be redundant.  The decorator itself is needed in Angular 10 however - it won't compile without it.

src/app/dataset/reportsimple/reportsimple.component.ts
andyjmaclean: Window.getSelection assumes there is a selection to be gotten.  The call to getSelection has been moved to an inline function.

src/app/dashboard/executionsgrid/gridrow/gridrow.component.ts
```diff
@@ -11,7 +11,9 @@ import { DatasetOverview, PluginExecutionOverview } from '../../../_models';
   styleUrls: ['./gridrow.component.scss']
 })
 export class GridrowComponent {
-  @ViewChild('childComponentTemplate') childComponentTemplate: TemplateRef<HTMLElement>;
+  @ViewChild('childComponentTemplate', { static: true }) childComponentTemplate: TemplateRef<
+    HTMLElement
+  >;
```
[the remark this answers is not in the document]
andyjmaclean: The migration tool added these changes.  Going to angular 9 a `static` declaration appeared for every `ViewChild` in the code base - either true or false.  Going to angular 10 the `static: false` declarations vanished (as that became the default)

The purpose is to make the child components available before ngOnViewInit - I was happy to let the compiler work it out for me.   Note that removing the declaration from `AppComponent` causes a unit test to fail  (cannot read property id from undefined...) - similar errors occur elsewhere.

I've made a further change to `AppComponent`, which no longer has to implement both `OnInit` and `OnAfterViewInit`, as what was an ambiguous state of `ViewChild` declarations on init is made determinate if `static:true` is present in the decorator.

Removing the `static: true` statement in FilterOptionComponent creates many errors of the type:

`Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: '[object Object]'.`

(timeouts can address this, but obviously this is ultra-hacky, so it's better to use leave the statement where it is)

It was possible, however, to remove `static: true` from the following places:
* GridRowComponent
* DepublicationRowComponent
* SortableGroupComponent
* ExecutionsDataGridComponent
* WorkflowFormFieldComponent
* WorkflowHeaderComponent

src/app/app.component.ts
```diff
@@ -23,7 +23,7 @@ export class AppComponent extends SubscriptionManager implements OnInit, AfterVi
   public loggedIn = false;
   modalConfirmId = 'confirm-cancellation-request';
 
-  @ViewChild(ModalConfirmComponent)
+  @ViewChild(ModalConfirmComponent, { static: true })
```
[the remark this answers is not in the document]
andyjmaclean: See below

src/app/dataset/preview/preview.component.spec.ts
```diff
@@ -309,7 +314,7 @@ describe('PreviewComponent', () => {
       expect(fixture.debugElement.queryAll(By.css('.view-sample-compared')).length).toBe(0);
 
       component.getXMLSamplesCompare(PluginType.NORMALIZATION, '123', true);
-      tick(1);
+      tick(5);
```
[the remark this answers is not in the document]
andyjmaclean: Reverted

src/app/dataset/reportsimple/reportsimple.component.ts
jochen-vermeulen: I'm not sure I understand your reply. :)

My point is that you don't need the inline function you created because all it does is check if there is a selection, and if there is, apply the operation to it. But we now have optional chaining for that (the `?` that is inserted in my suggestion). Wouldn't that apply here?

PR #505, opened 2020-11-27
Remove cancel option for DEPUBLISH

andyjmaclean:


src/app/dashboard/ongoingexecutions/ongoingexecutions.component.ts
```diff
@@ -45,6 +46,18 @@ export class OngoingexecutionsComponent implements OnInit {
...
+  /*  @param {WorkflowExecution} ongoing - the execution
+  */
+  canCancelWorkflow(ongoing: WorkflowExecution): boolean {
+    return (
+      [`${PluginType.PUBLISH}`, `${PluginType.DEPUBLISH}`].indexOf(
+        ongoing.currentPlugin!.pluginType
+      ) == -1 && !ongoing.cancelling
+    );
+  }
+
```
jochen-vermeulen, line 60: This function is not used (except in test code). I think it needs to be called from the html somewhere?

src/app/dataset/actionbar/actionbar.component.ts
```diff
@@ -142,6 +144,16 @@ export class ActionbarComponent {
...
   }
 
+  /** canCancelWorkflow
+  /*  calculate if the workflow can be cancelled
+  */
+  canCancelWorkflow(): boolean {
+    return (
+      [`${PluginType.PUBLISH}`, `${PluginType.DEPUBLISH}`].indexOf(`${this.currentPluginName}`) ==
+        -1 && !this.isCompleted
+    );
```
jochen-vermeulen, line 154: Seems like duplicate code - you added this check twice. Is there a possibility to have one source of truth somewhere?

src/app/dashboard/ongoingexecutions/ongoingexecutions.component.ts
andyjmaclean: Fixed

src/app/dataset/actionbar/actionbar.component.ts
andyjmaclean: It was similar - so making the helper wasn't painless.

PR #509, opened 2021-02-01
Reconfigure as Angular workspace

andyjmaclean:


.travis.yml
```diff
@@ -16,5 +16,8 @@ addons:
 before_install:
 - export TZ=Europe/Amsterdam
 
+# before_script:
+# - cp src/environments/apisettings-test.ts src/environments/apisettings.ts
```
jochen-vermeulen, line 20: Should be removed?

angular.json
```diff
@@ -4,25 +4,35 @@
   "newProjectRoot": "projects",
   "projects": {
     "metis": {
-      "root": "",
-      "sourceRoot": "src",
+      "root": "projects/metis",
```
jochen-vermeulen, line 7: Now that you're setting this root, should you not define all paths below relative to this root?

package.json
```diff
@@ -5,27 +5,46 @@
...
-    "start": "ng serve",
-    "build": "ng build",
-    "prettier": "prettier --write '{src,cypress,test-data}/**/*.{ts,js,json,scss,html}' '*.{js,json}'",
-    "lint": "eslint --ext .js,.ts src/app --fix",
+    "start-metis": "ng serve metis",
+    "start": "npm run start-metis",
+    "build-metis": "ng build metis",
+    "build": "npm run build-metis",
+    "prettier-metis": "prettier --write '{projects/metis/src,projects/metis/cypress,projects/metis/test-data}/**/*.{ts,js,json,scss,html}' '*.{js,json}'",
+    "prettier": "npm run prettier-metis",
```
jochen-vermeulen, line 13: If there is a second project, would it also be triggered if you run prettier? Or would you always have to make multiple calls to make sure you run prettier on everything? I think there should be a call for all projects (if you work on multiple), but this is perhaps not always possible (e.g. running start). What's the idea here?

andyjmaclean: I'll address that in the follow up ticket once another project has been added.  Prettier doesn't take long to run, so the default action probably should be to run it on everything.  As configured at present it will only operate on the files under `projects/metis`

.travis.yml
andyjmaclean: Yes

angular.json
andyjmaclean: Those were my thoughts at first, but it only works when configured as it is here.

package.json
jochen-vermeulen: OK, let's just keep that in mind.

angular.json
jochen-vermeulen: Odd... but I believe you :)

PR #528, opened 2021-03-22
Extracted ternary operators

andyjmaclean:


projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -302,7 +312,11 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
     }
 
     const validateTimer = timer(10).subscribe(() => {
-      this.workflowStepAllowed(this.inputFields ? this.inputFields.toArray() : undefined);
+      if (this.inputFields) {
+        this.workflowStepAllowed(this.inputFields.toArray());
+      } else {
+        this.workflowStepAllowed(undefined);
+      }
```
jochen-vermeulen, line 319: For consistency, perhaps do this (lines 315-319) the same way as you did it below (lines 331-335)? I think that way is better, don't you?

projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -382,9 +402,11 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
     }
     // parameters for link-checking
     if (enabledPluginMetadata.pluginType === 'LINK_CHECKING') {
-      this.workflowForm.controls.performSampling.setValue(
-        enabledPluginMetadata.performSampling ? 'true' : 'false'
-      );
+      let value = 'false';
+      if (enabledPluginMetadata.performSampling) {
+        value = 'true';
+      }
```
jochen-vermeulen, line 408: I think you can replace these four lines by something like this?
```
let value = String(enabledPluginMetadata.performSampling);
```

projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -498,10 +523,11 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
...
         } else {
-          conf.parameterFields = conf.parameterFields ? conf.parameterFields : [];
-          plugins.push(
-            this.formatFormValue(conf.label as PluginType, conf.parameterFields, enabled)
-          );
+          let parameterFields: ParameterField = [];
+          if (conf.parameterFields) {
+            parameterFields = conf.parameterFields;
+          }
+          plugins.push(this.formatFormValue(conf.label as PluginType, parameterFields, enabled));
```
jochen-vermeulen, line 530: The old version had the side effect of ensuring that `conf.parameterFields` is always defined. Is that effect no longer needed?

projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -149,22 +149,27 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
...
   }
 
   /** setHighlightedField
    * marks header orb as highlighted if it's the topmost in the viewport
    */
   setHighlightedField(fields: Array<WorkflowFormFieldComponent>, headerEl?: HTMLElement): void {
-    const headerHeight = 77 + (headerEl ? headerEl.offsetHeight : 0);
+    const headerHeight = 77;
+    if (headerEl) {
+      headerHeight + headerEl.offsetHeight;
```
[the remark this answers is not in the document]
andyjmaclean: That was incredibly careless

projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -173,7 +178,13 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
       if (!scorePositive && scoreA + scoreB > 0) {
         scorePositive = true;
       }
-      return scoreA === scoreB ? 0 : scoreA > scoreB ? -1 : 1;
+      if (scoreA === scoreB) {
+        return -1;
```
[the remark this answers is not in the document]
andyjmaclean: Maybe that would conform to a generic convention but the idea here was to remove the ternary, not rewrite the logic!

projects/metis/src/app/dataset/workflow/workflow.component.ts
andyjmaclean: Yes

projects/metis/src/app/dataset/workflow/workflow.component.ts
andyjmaclean: Yes

projects/metis/src/app/dataset/workflow/workflow.component.ts
andyjmaclean: According to my testing the effect was never needed!

projects/metis/src/app/dataset/workflow/workflow.component.ts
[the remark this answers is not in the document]
jochen-vermeulen: But you are rewriting the logic. In the old logic you had `0` if `scoreA === scoreB`, and in your new logic this would result in `-1`.

projects/metis/src/app/dataset/workflow/workflow.component.ts
jochen-vermeulen: It would seem not to be a dangerous change.

projects/metis/src/app/dataset/workflow/workflow.component.ts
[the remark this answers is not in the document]
andyjmaclean: Yikes.  Fixed.

PR #529, opened 2021-03-22
Use real service for countries and languages

andyjmaclean:


projects/sandbox/src/app/wizard/wizard.component.html
```diff
@@ -74,15 +74,17 @@
...
         </select>
       </div>
 
       <div *ngIf="currentStepIndex === 1" class="form-group">
         <label for="language" i18n="language label">Language</label>
         <select id="language" formControlName="language">
-          <option *ngFor="let option of languageList" [ngValue]="option.code">{{
-            option.name
+          <option *ngFor="let option of languageList" [ngValue]="option.xmlValue">{{
+            option.xmlValue
```
jochen-vermeulen, line 87: I think for languages the value that's sent back (i.e. the option's value attribute) should be equal to the language's or country's name property (and not the xmlValue property). One is for display, the other is for internal use. They just happen to be the same (apart from whether they are lower case or upper case) but this may change.

PR #537, opened 2021-03-30
Rename app files to .mjs for e2e tests

andyjmaclean:


sonar-project.properties
```diff
@@ -1,16 +1,5 @@
...
-# for angular app-etc selectors
-sonar.issue.ignore.multicriteria.e2.ruleKey=scss:unknown-type-selector
-sonar.issue.ignore.multicriteria.e2.resourceKey=**/*.scss
-# for angular :host selectors
-sonar.issue.ignore.multicriteria.e3.ruleKey=scss:unknown-pseudo
-sonar.issue.ignore.multicriteria.e3.resourceKey=**/*.scss
-
+sonar.exclusions=**/_mocked/**.ts
+sonar.projectKey=Metis-UI
+sonar.javascript.lcov.reportPaths=projects/metis/coverage/Chrome\ Headless\ 89.0.4389.90\ \(Linux\ x86_64\)/lcov.info,projects/sandbox/coverage/Chrome\ Headless\ 89.0.4389.90\ \(Linux\ x86_64\)/lcov.info,projects/shared/coverage/Chrome\ Headless\ 89.0.4389.90\ \(Linux\ x86_64\)/lcov.info
```
jochen-vermeulen, line 5: What's this number? 89.0.4389.90? It really looks like a Chrome version number. I don't think that's sustainable to have that hardcoded, as it will be outdated in a matter of weeks. In fact, as we speak, I think my chrome is currently at 89.0.4389.114 already.

package.json
```diff
@@ -36,7 +35,8 @@
     "test:ci-sandbox": "npm run test:ci-any --project=sandbox",
     "test:ci-shared": "npm run test:ci-any --project=shared",
     "test:ci": "npm run test:ci-sandbox && npm run test:ci-shared && npm run test:ci-metis",
-    "test:e2e-any": "npm run tsc-test-data --project=${npm_config_project} && ((node --experimental-specifier-resolution=node --use_strict ./projects/${npm_config_project}/dist/out-tsc/projects/${npm_config_project}/test-data/index.js) & (npm run cypress:ci-any --project=${npm_config_project}; fuser -k 3000/tcp))",
+    "prep-test-data": "node tools/test-data-prep/rename-files-to-mjs.mjs projects/${npm_config_project}/dist/out-tsc",
```
jochen-vermeulen, line 38: I think we should add some pretty detailed documentation to describe what we're doing here and why.

sonar-project.properties
andyjmaclean: Yes, that's what it is.  The numbers will change - I included this as an indicator for developers (this file is used by the developer who wants to run a local scan on the project - not by any build) but the developers have access to coverage information elsewhere.  For those reasons I've removed it.

package.json
andyjmaclean: Done (added comment above script)

PR #538, opened 2021-03-31
Improved form validation

andyjmaclean:


projects/sandbox/src/app/wizard/wizard.component.ts
```diff
@@ -96,18 +96,37 @@ export class WizardComponent extends DataPollingComponent {
...
+    const sVal = `${val}`;
     if (val) {
-      const isWhitespace = val.length > 0 && val.trim().length === 0;
-      if (isWhitespace) {
+      if (isNaN(val) || sVal.length === 0 || sVal.length !== `${parseInt(val)}`.length) {
+        return { invalid: true };
+      }
+    }
+    return null;
+  }
```
jochen-vermeulen, line 115: This seems overly complicated and also doesn't do exactly what we want, I think? What about `-4`? Or perhaps weird scientific notation (`5e4` or something like that)?

In my mind it's best to do it like below in the name validator. Use a regex to make sure that it matches only digits.

projects/sandbox/src/app/wizard/wizard.component.ts
```diff
@@ -96,18 +96,37 @@ export class WizardComponent extends DataPollingComponent {
...
+   *
+   * form validator implementation for dataset name field
+   *
+   * @param { FormControl } control - the control to validate
+   * @returns null or a code-keyed boolean
+   **/
+  validateDatasetName(control: FormControl): { [key: string]: boolean } | null {
+    const val = control.value;
+    if (val) {
+      const matches = val.match(/[a-zA-Z0-9\\_\\]+/);
```
jochen-vermeulen, line 128: I think there is something wrong here? This allows backslash twice, right? I think you could just put `/[a-zA-Z0-9_]+/`, right?

projects/sandbox/src/app/wizard/wizard.component.ts
andyjmaclean: Fixed

projects/sandbox/src/app/wizard/wizard.component.ts
andyjmaclean: Yes

PR #541, opened 2021-04-01
Use separate statements for assignment

andyjmaclean:


projects/metis/src/app/_helpers/cache.ts
```diff
@@ -13,23 +13,26 @@ export class SingleCache<Value> {
     if (this.observable && !refresh) {
       return this.observable;
     }
-    const observable = (this.observable = this.sourceFn().pipe(
+    const observable = this.sourceFn().pipe(
```
jochen-vermeulen, line 16: Why not directly assign it to this.observable? Doesn't look like we need the local variable.

projects/metis/src/app/_helpers/cache.ts
```diff
@@ -53,23 +56,24 @@ export class KeyedCache<Value> {
     if (o && !refresh) {
       return o;
     }
-    const observable = (this.observableByKey[key] = this.sourceFn(key).pipe(
+    const observable = this.sourceFn(key).pipe(
```
jochen-vermeulen, line 59: Same here.

projects/metis/src/app/_helpers/cache.ts
jochen-vermeulen: We agreed that this is not possible, and added comments (and better unit tests) to that effect.

PR #548, opened 2021-04-15
Comment out incremental harvest functionality

andyjmaclean:


projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -385,10 +387,10 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
...
         this.workflowForm.controls.metadataFormat.setValue(thisWorkflow.metadataFormat);
-        this.workflowForm.controls.incrementalHarvest.setValue(thisWorkflow.incrementalHarvest);
+        //this.workflowForm.controls.incrementalHarvest.setValue(thisWorkflow.incrementalHarvest);
       }
     }
-    this.enableIncrementalHarvestingFieldIfAvailable(workflow);
+    //this.enableIncrementalHarvestingFieldIfAvailable(workflow);
   }
 
   /** extractPluginParamsExtra
```
jochen-vermeulen, line 396: I think in this file we also need to remove the reference to `ParameterFieldName.incrementalHarvest` in line 499 (method `formatFormValues()`), just to be safe. Does that make sense?

andyjmaclean: Done!

PR #549, opened 2021-04-16
ref -jade in svgs / colour update / dead code removal / grid tweaks

andyjmaclean:


projects/metis/src/app/dataset/lastexecution/lastexecution.component.ts
```diff
@@ -48,8 +49,19 @@ export class LastExecutionComponent {
...
   }
 
+  /** getPluginHighlighted
+  /* template utility to determine if row should be highlighted
+  /* @param { PluginExecution } plugin - the plugin to evaluate
+  /* @return boolean
+  */
+  getPluginHighlighted(plugin: PluginExecution): boolean {
+    return plugin.pluginStatus === PluginStatus.RUNNING;
+  }
```
jochen-vermeulen, line 59: I think, as a result from an earlier comment, this method would no longer be necessary.

projects/metis/src/app/dataset/history/history.component.html
```diff
@@ -6,42 +6,52 @@
...
-    {{ 'endTime' | translate }}
-  </div>
-  <div class="table-grid-header"></div>
-
-  <ng-container *ngFor="let wpe of allExecutions; trackBy: byId">
-    <div
-      *ngIf="!wpe.pluginExecution && wpe.execution; let execution"
-      class="sub-header workflow-head"
-    >
+  <ng-container *ngFor="let wpe of allExecutions; trackBy: byId; let i = index">
```
jochen-vermeulen, line 9: I think this container and everything in it is just unnecessarily complicated. You have this strange structure called allExecutions that tries to flatten the hierarchy of workflow executions and plugin executions inside them. I don't see the need to do that.

Why not have a list of workflow executions, and then you have the first ng-container below, and then you loop through the plugins in the workflow to have instances of `app-executions-grid-data`? That way the code makes much more sense, you don't need an awkward data structure and, even better, you can then compute the index as you normally would by taking the modulo, instead of having to use this awkward `rowIsStripe` function.

projects/metis/src/app/dataset/executions-data-grid/executions-data-grid.component.ts
```diff
@@ -19,7 +19,8 @@ import {
 })
 export class ExecutionsDataGridComponent {
   @Input() hasMore?: boolean;
-
+  @Input() applyHighlight?: boolean;
+  @Input() applyStripe?: boolean;
```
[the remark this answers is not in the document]
andyjmaclean: I agree on `applyHighlight` - it could be contained within the component and doesn't need the `Input` decorator, but I do think `applyStripe` is a more descriptive name than `indexInWorkflow`.

projects/metis/src/app/dataset/lastexecution/lastexecution.component.ts
andyjmaclean: I'll move it to a setter for the plugin `Input`

projects/metis/src/app/dataset/executions-data-grid/executions-data-grid.component.ts
```diff
@@ -19,7 +19,8 @@ import {
 })
 export class ExecutionsDataGridComponent {
   @Input() hasMore?: boolean;
```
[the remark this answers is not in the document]
andyjmaclean: No.  I've removed it

projects/metis/src/app/dataset/executions-data-grid/executions-data-grid.component.ts
```diff
@@ -40,13 +45,13 @@ export class ExecutionsDataGridComponent {
...
-  goToPreview(execution: WorkflowExecution, pluginExecution: PluginExecution): void {
+  goToPreview(executionId: string, pluginExecution: PluginExecution): void {
     const previewFilters: PreviewFilters = {
       baseFilter: {
-        executionId: execution.id,
+        executionId: executionId,
         pluginType: pluginExecution.pluginType
       },
-      baseStartedDate: execution.startedDate
+      baseStartedDate: pluginExecution.startedDate
```
jochen-vermeulen, line 54: Are you sure this change is correct? The starting date of the workflow execution is not the same as the starting date of the plugin.

projects/metis/src/app/dataset/history/history.component.ts
```diff
@@ -74,13 +71,6 @@ export class HistoryComponent extends SubscriptionManager {
...
-  /** resetTemplateRowIndex
-  /* Resets templateRowIndex to 0
-  */
-  resetTemplateRowIndex(): void {
-    this.templateRowIndex = 0;
-  }
-
   /** rowIsStripe
   /* Increments templateRowIndex variable and returns its modulus of 2
   /* @return boolean
```
jochen-vermeulen: The function `rowIsStripe()` is no longer required, and should be removed, I think? The same for `this.templateRowIndex`.

PR #552, opened 2021-04-28
Add deleted test data

andyjmaclean:


projects/metis/test-data/factory/factory.ts
```diff
@@ -183,6 +184,7 @@ function runWorkflow(workflow: WorkflowX, executionId: string): WorkflowExecutio
 
       const prc = {
         numExpected: wConf.expectedRecords,
+        numDeleted: pluginExecutionIsHarvest(pe) && wConf.deletedRecords ? wConf.deletedRecords : 0,
```
jochen-vermeulen, line 187: I'm not convinced that deleted records are confined only to harvest jobs. In fact, I quite think that they will at least also appear in preview and publish jobs. Would that make a difference here?

andyjmaclean: It's better to represent real data so I've made a change.

PR #554, opened 2021-04-30
Simplify validation / tests

andyjmaclean:


projects/metis/src/app/dataset/workflow/workflow-header/workflow-header.component.ts
```diff
@@ -41,6 +41,15 @@ export class WorkflowHeaderComponent implements AfterViewInit {
       ctrl.setValue(!ctrl.value);
       this.workflowForm.markAsDirty();
     }
+    if (plugin === 'pluginHARVEST') {
+      const ctrl = this.workflowForm.get('pluginType') as FormControl;
```
jochen-vermeulen, line 45: You must be able to find this string value 'pluginType' in the workflow settings somewhere, right? It should be a constant or otherwise obtainable from the workflow config?

projects/metis/src/app/dataset/workflow/workflow.component.ts
```diff
@@ -329,11 +326,6 @@ export class WorkflowComponent extends SubscriptionManager implements OnInit {
   updateRequired(): void {
     this.subs.push(
       this.workflowForm.valueChanges.subscribe(() => {
-        let fieldsArray: Array<WorkflowFormFieldComponent> | undefined = undefined;
-        if (this.inputFields) {
-          fieldsArray = this.inputFields.toArray();
-        }
-        this.workflowStepAllowed(fieldsArray);
```
jochen-vermeulen, line 336: Shouldn't this be replaced by a call to hasGapInSequence? So that it has the required side-effect?

projects/metis/src/app/dataset/workflow/workflow-header/workflow-header.component.ts
andyjmaclean: It's defined [here](https://github.com/europeana/metis-ui/blob/develop/projects/metis/src/app/_models/workflow-form.ts#L23-L33) as a type, and types "cannot be used as a value".

projects/metis/src/app/dataset/workflow/workflow.component.ts
andyjmaclean: It's no longer needed, and in any case it was firing after validation had already occurred (which is why the gap in sequence error showed only after TWO gaps were created)

projects/metis/src/app/dataset/workflow/workflow-form-field/workflow-form-field.component.ts
```diff
@@ -42,11 +42,4 @@ export class WorkflowFormFieldComponent {
     }
     return !(this.workflowForm.get(this.conf.name) as FormControl).value;
   }
-
-  /** onFieldChanged
-  /* emits fieldChanged event
-  */
-  onFieldChanged(fieldName: string): void {
-    this.fieldChanged.emit(fieldName);
```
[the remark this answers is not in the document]
andyjmaclean: I've removed both

projects/metis/src/app/dataset/workflow/workflow-header/workflow-header.component.ts
jochen-vermeulen: I understand, but isn't it better in this case to have `this.workflowForm.get(ParameterFieldName.pluginType)`? Or wouldn't that work for some reason?

PR #558, opened 2021-05-10
Button styles

andyjmaclean:


projects/metis/src/app/dataset/executions-data-grid/executions-data-grid.component.ts
```diff
@@ -63,6 +63,7 @@ export class ExecutionsDataGridComponent {
   */
   harvestIsIncremental(pluginExecution: PluginExecution): boolean {
     return (
+      pluginExecution.pluginMetadata &&
```
jochen-vermeulen, line 66: There is a model issue here. Whatever comes out of the endpoint
`/orchestrator/workflows/executions/dataset/{datasetid}`
has a different model than some of the other requests and can never have a field called pluginMetadata.
So I think in this case there is some mixup of model objects. 
And obviously that means that this method will always return false.

jochen-vermeulen (changes requested):

projects/metis/src/app/dataset/executions-data-grid/executions-data-grid.component.ts
andyjmaclean: This mis-modeling was found in a lot of the test data, including (now deleted) cypress fixtures that haven't been used in over a year.  I've added `isIncremental` to `WorkflowExecution` (as per the test-data server).  In addition I've removed the `renameStatus` pipe and used `ngSwitch` to list the statuses in an i18n-annotatable way.

projects/metis/src/app/dataset/actionbar/actionbar.component.spec.ts
```diff
@@ -22,6 +22,7 @@ describe('ActionbarComponent', () => {
       declarations: [
         ActionbarComponent,
         createMockPipe('translate'),
+        createMockPipe('renameStatus'),
```
jochen-vermeulen, line 25: This mock pipe is no longer required, right?

PR #563, opened 2021-05-18
Added tests / code tidy

andyjmaclean:


projects/metis/src/app/dataset/datasetlog/datasetlog.component.ts
```diff
@@ -92,10 +92,6 @@ export class DatasetlogComponent extends DataPollingComponent implements OnInit
       () => {
         return of(this.getProcessedCount()).pipe(
           map((val) => {
-            if (isPluginCompleted(this.showPluginLog) && this.subs.length > 0) {
-              this.cleanup();
-              return 0;
-            }
```
jochen-vermeulen, line 98: Are you sure this should be removed? If the task is complete, you should stop polling, right?

PR #626, opened 2022-03-08
Feat/met 4211 add accessibility checks to sandbox

andyjmaclean:


projects/metis/cypress.json
```diff
@@ -1,6 +1,6 @@
 {
   "baseUrl": "http://localhost:4280",
-  "blacklistHosts": ["*fonts.googleapis.com", "*fonts.gstatic.com"],
+  "blockHosts": ["*fonts.googleapis.com", "*fonts.gstatic.com"],
```
adolfopeixinho, line 3: Political Correctness got to you :)

projects/sandbox/src/app/app.component.html
```diff
@@ -3,7 +3,7 @@
     <div class="header-container">
       <div class="metis-logo">
         <span class="phase-flag beta" tabindex="-1">BETA</span>
-        <a href="/" class="logo"></a>
+        <a href="/" aria-label="Europeana Metis Sandbox" class="logo"></a>
```
adolfopeixinho, line 6: Does this fix the issue with clicking on Europeana logo not working?

adolfopeixinho (approved):

projects/sandbox/src/app/app.component.html
andyjmaclean: Making the url a slash rather than a Router command would have fixed that specific issue, but the problem was deeper, and that bug was due to Angular.Location losing track of the current url due to some logic I fixed as part of MET-4266.

The Aria-Label is needed to stop Axe complaining about the link not containing any text.

PR #654, opened 2022-06-29
MET-4619 Replace RXJS tap signature / publishLast with non-deprecated …

andyjmaclean:
…equivalents

### DIFF
--- a/projects/metis/src/app/_helpers/cache.ts
+++ b/projects/metis/src/app/_helpers/cache.ts
@@ -1,5 +1,5 @@
-import { Observable, of } from 'rxjs';
-import { shareReplay, tap } from 'rxjs/operators';
+import { AsyncSubject, connectable, Observable, of } from 'rxjs';
+import { tap } from 'rxjs/operators';
 
 export class SingleCache<Value> {
   private observable?: Observable<Value>;
@@ -15,15 +15,21 @@ export class SingleCache<Value> {
       return this.observable;
     }
 
-    const observable = this.sourceFn().pipe(
-      tap({
-        error: () => this.clear()
-      }),
-      shareReplay(1)
+    // Create new observable.
+    const observable = connectable(
+      this.sourceFn().pipe(
+        tap({
+          error: () => this.clear()
+        })
+      ),
+      {
+        connector: () => new AsyncSubject<Value>(),
+        resetOnDisconnect: false
+      }
     );
-
     // Return local variable, as this.observable might be cleared by the connect call.
     this.observable = observable;
+    observable.connect();
     return observable;
   }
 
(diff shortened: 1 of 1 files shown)


projects/metis/src/app/_helpers/cache.ts
```diff
@@ -15,19 +15,15 @@ export class SingleCache<Value> {
...
-    // Create new observable.
     const observable = this.sourceFn().pipe(
-      tap(
-        (_) => undefined,
-        (_) => this.clear(),
-        () => undefined
-      ),
-      publishLast()
-    ) as ConnectableObservable<Value>;
+      tap({
```
jeortizquan, line 19: I found this documentation on how to upgrade ```publishLast``` to a combination of ```connectable``` and ```AsyncSubject``` have a look on the ```publishLast``` section

https://rxjs.dev/deprecations/multicasting

and can this I think can be the upgrade. I just tested in my local environment. Can you validate if this does the same as before? I understand now you are using RxJS 7.5.5. This could apply to the ```SingleCache``` and ```KeyedCached``` classes

```
import { AsyncSubject, connectable, Observable, of } from 'rxjs';
import { tap } from 'rxjs/operators';

export class SingleCache<Value> {
  private observable?: Observable<Value>;

  constructor(private readonly sourceFn: () => Observable<Value>) {}

  /** get
  /* accessor for the observable variable
  */
  public get(refresh = false): Observable<Value> {
    // If already cached, return.
    if (this.observable && !refresh) {
      return this.observable;
    }

    // Create new observable.
    const observable = connectable(this.sourceFn().pipe(
        tap(
          (_) => undefined,
          (_) => this.clear(),
          () => undefined
        )), {
          connector: () => new AsyncSubject<Value>(),
        resetOnDisconnect: false
      });
    // Return local variable, as this.observable might be cleared by the connect call.
    this.observable = observable;
    observable.connect();
    return observable;
  }

  /** peek
  /* return the observable or undefined as an observable
  */
  public peek(): Observable<Value | undefined> {
    return this.observable || of(void 0);
  }

  /** clear
  /* sets the observable to undefined
  */
  public clear(): void {
    this.observable = undefined;
  }
}

export class KeyedCache<Value> {
  private readonly observableByKey: { [key: string]: Observable<Value> | undefined } = {};

  constructor(private readonly sourceFn: (key: string) => Observable<Value>) {}

  /** get
  /* accessor for the observable variable
  */
  public get(key: string, refresh = false): Observable<Value> {
    // If already cached, return.
    const o = this.observableByKey[key];
    if (o && !refresh) {
      return o;
    }

    // Create new observable.
    const observable = connectable(this.sourceFn(key).pipe(
      tap(
        (_) => undefined,
        (_) => this.clear(key),
        () => undefined
      )), {
      connector: () => new AsyncSubject<Value>(),
      resetOnDisconnect: false
    });

    // Return local variable, as this.observableByKey[key] might be cleared by the connect call.
    this.observableByKey[key] = observable;
    observable.connect();
    return observable;
  }

  /** peek
  /* return the observable or undefined as an observable
  */
  public peek(key: string): Observable<Value | undefined> {
    return this.observableByKey[key] || of(void 0);
  }

  /** clear
  /* sets the observable to undefined
  */
  public clear(key: string): void {
    delete this.observableByKey[key];
  }
}
```

jeortizquan (commented):
I placed some comments. I found this suggestion, in order to have the same behaviour as it was before. What do you think?

projects/metis/src/app/_helpers/cache.ts
andyjmaclean: I think you're right.  I took the AsyncSubject solution - with the updated "tap" syntax

jeortizquan (approved):

PR #703, opened 2022-10-11
Feat/met 4823 metis show throttle level

andyjmaclean:


jeortizquan (approved):
It shows on the history

image

and on the main processing tab.
image

PR #705, opened 2022-11-07
Feat/met 4935 problem pattern navigation

andyjmaclean:


projects/sandbox/src/app/problem-viewer/problem-viewer.component.html
```diff
@@ -236,17 +263,19 @@
...
-          tmpRecordAnalysisList;
-          context: {
-            analyses: problemPattern.recordAnalysisList,
-            problemTitle: problemPattern.problemPatternDescription.problemPatternTitle,
-            datasetId: datasetId,
-            ppId: problemPattern.problemPatternDescription.problemPatternId
-          }
-        "
-      ></ng-content>
+      <ng-container *ngIf="!problemPattern.closed">
```
JoanaCMS, line 266: Is there a boolean method `open`? It would be easier to read if it were `ngIf="problemPattern.open"`

andyjmaclean: There is no boolean method - just the "closed" field in the augmented model.

andyjmaclean: > Is there a boolean method `open`? It would be easier to read if it were `ngIf="problemPattern.open"`

There is no boolean method - just the "closed" field in the (augmented) model - and that's been there since problem patterns were first introduced!

JoanaCMS (approved):

PR #715, opened 2022-12-05
MET-4965 Add Theme Switching

andyjmaclean:


projects/sandbox/src/app/footer/footer.component.scss
```diff
@@ -44,6 +44,7 @@
```
JoanaCMS, line 44: I saw that the changes done to the footer are not many, at least in this PR.
But I noticed that the footer itself does change between each theme, is that the desired behavior?
Also, a personal opinion, I don't think the colors of the footer go together with the whole color scheme in both themes of sandbox, and perhaps it's too big. Was this design agreed previously?

projects/sandbox/src/app/home/home.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="home">
-  <div class="home-header">
-    <h1>Welcome to<br />Metis Sandbox!</h1>
-    <div class="circle-container-outer">
-      <ul class="circle-container">
+  <div class="wizard-head">
```
JoanaCMS, line 2: I've seen the word `wizard` being used in multiple places, what is this referring to?
For me it is not clear/obvious what this is meant to represent

andyjmaclean: > I've seen the word `wizard` being used in multiple places, what is this referring to? For me it is not clear/obvious what this is meant to represent

This is legacy nomenklature.  I've created [this ticket](https://europeana.atlassian.net/browse/MET-5008) to remove it

projects/sandbox/src/app/footer/footer.component.scss
andyjmaclean: For now the footer contains only the "sub-footer" of the portal, so it varies from the design in content, which has still to be decided (and which will better fill it out).  More redesign iterations are anticipated.

Ideally the theme switch would load the old footer from the shared library, but it's a very lightweight and temporary mechanism that conditionally includes css but which doesn't alter markup structure.  By the time we release issues of colour clash should be gone!

JoanaCMS (approved):

PR #729, opened 2023-02-07
MET-5082 Added Basic Responsivity

andyjmaclean:


jeortizquan (commented):
**In Chrome** 
image
* in the old version the main content is pushed down if the hamburger button is pressed.
* in the new version the main content remains in place unchanged, and also the menu bar fills about 50% of the screen. 

**In Firefox**
image
* in the old version the menu fills 100% when hamburger button is pressed and the main content is pushed down a little bit.
* in the new version the main content fills 100% and the menu fills 100% and removes the rounded corners.

Comments:
* Should we make at least chrome and firefox look alike?
* if the viewport is shrinked, and the menu is displayed should the main content pushed down and the menu be displayed in the top of it? like this

image

jeortizquan (approved):
We discover that In the default zoom mode of google chrome we cannot make the window smaller to trigger the breaking point. However the user will be able to see the menu even if scrolls down.

PR #730, opened 2023-02-08
MET-4989 Remove P1 links

andyjmaclean:


jeortizquan (approved):
The links are no longer clickable for P1
verified with cypress and local branch...

PR #731, opened 2023-02-09
MET-5086 Bind scroll-preventing body class to modal open / close

andyjmaclean:


projects/metis/src/app/dataset/reportsimple/reportsimple.component.html
```diff
@@ -1,102 +1,107 @@
...
-      Report
-      <span *ngIf="reportRequest.pluginType; let plugin">({{ plugin | renameWorkflow }})</span
-      ><a class="btn-close" (click)="close()"></a>
-      <a
-        class="btn-right"
-        (click)="copyReport()"
-        *ngIf="reportRequest.errors || reportRequest.message"
-        >Copy</a
-      >
+<ng-template #tmpBtnCopy>
```
JoanaCMS, line 1: The name of this template `tmpBtnCopy` is not clear. Is it meant to be `Template Button Copy`?
Could you change it to a name that is easier to read/understand?

projects/metis/src/app/dataset/reportsimple/reportsimple.component.html
```diff
@@ -1,102 +1,107 @@
...
-                      </span>
-                    </li>
-                    <li
-                      class="download-error omit-from-copy"
-                      *ngIf="
-                        sub === 'identifier' && reportErrors[r][rk][subrk].downloadError;
-                        let downloadError
+    <div *ngIf="reportRequest.errors; let reportErrors">
+      <ul *ngFor="let r of reportKeys(reportErrors)">
+        <li *ngFor="let rk of reportKeys(reportErrors[r])">
```
JoanaCMS, line 42: What are these variable `r` and `rk` meant to be?

andyjmaclean: The variables `r` (report) and `rk` (report key) were already in place and not introduced as part of this PR.

I wanted to do the minimum amount of refactoring to get all modal behaviour configured in the one place, which in the case of this file was about changing the wrapping element to a modal dialog instance.

But I wonder if having longer and explicit names would actually make the file more readable?  For me expressions can become less readable when they contain both the singular and the plural (`reportKey` / `reportKeys` etc.), especially in lines 41-56 of this file, which become far longer with the unabbreviated names (especially when `subrk` is introduced!)

projects/metis/src/app/dataset/reportsimple/reportsimple.component.html
andyjmaclean: The conventions `tmp` and `btn` are used widely in the project, so this is consistent.

jeortizquan: In my opinion maybe button copy needs to have a more representative/descriptive name

projects/metis/src/app/dataset/reportsimple/reportsimple.component.html
jeortizquan: the names `report` and `reportKey` made sense in this context, me as a external reader of the code I was able to understand this part of the code.

jeortizquan (commented):
I tested the changes locally and it works as intended.
image
just have a look on the comments and suggestions.

projects/shared/src/lib/modal-confirm/modal-confirm.component.ts
```diff
@@ -10,15 +19,18 @@ import { ModalConfirmService } from '../_services/modal-confirm.service';
...
   @Input() isSmall = true;
+  @Input() templateHeadContent?: TemplateRef<HTMLElement>;
   @ViewChild('modalWrapper', { static: false }) modalWrapper: ElementRef;
 
   subConfirmResponse: Subject<boolean>;
   show = false;
+  bodyClassOpen = 'modal-open';
 
-  constructor(private readonly modalConfirms: ModalConfirmService) {
+  constructor(private readonly modalConfirms: ModalConfirmService, private renderer: Renderer2) {
```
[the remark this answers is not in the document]
jeortizquan: As we discussed yesterday, I just realize is an angular name :/ so nevermind.

projects/metis/cypress/e2e/depublications.cy.ts
```diff
@@ -67,7 +67,7 @@ context('metis-ui', () => {
           .scrollIntoView()
           .click(force);
         cy.get('[name=recordIds]').type(entries);
-        cy.get('.submit-form').click();
+        cy.get('.modal [type="button"]').click();
```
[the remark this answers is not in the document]
andyjmaclean: A variable is needed only if the value is reused but this selector is used only once, and I'd only put it up at the top if a variable was being used by multiple tests - this keeps things easy to read!

projects/shared/src/lib/_mocked/mocked-renderer-2.ts
```diff
@@ -0,0 +1,8 @@
+export class MockRenderer2 {
```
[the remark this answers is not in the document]
andyjmaclean: I'm following a convention with this name, which is to prefix the name of the thing being mocked with the word "Mock" (and the filename prefix 'mocked-').  Is it a good idea to break this convention?  The person reading the code would have to check the `providers` block in the `spec.ts` to confirm the actual name of the thing that was being mocked, if it were to be changed to that it could no longer be deduced.

projects/metis/src/app/dataset/reportsimple/reportsimple.component.html
andyjmaclean: I've renamed it to `tmpBtnCopyReport`.  I think that encapsulates all that a name can.

shared-styles/assets/sass/pandora/components/_modal.scss
```diff
@@ -1,11 +1,16 @@
...
+  overflow-y: hidden;
+}
+
 .modal-wrapper {
   background: $opacity-black;
   height: 100vh;
   left: 0;
   position: fixed;
   top: 0;
   width: 100%; // 100vw leaves uneven spacing
```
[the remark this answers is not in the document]
andyjmaclean: Done

projects/shared/src/lib/_mocked/mocked-renderer-2.ts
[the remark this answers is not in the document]
jeortizquan: I get it after the explanation you gave me about this object make sense to be named `MockRenderer2`

jeortizquan (approved):
The functionality is there.
image

PR #733, opened 2023-02-15
MET-5094 Use Renderer2 to Switch Theme

andyjmaclean:


jeortizquan (approved):
The switching of the theme works.
new theme
image
old theme
image

the swtiching feature continues working.

PR #735, opened 2023-02-23
MET-5047 Sandbox K8s

andyjmaclean:


projects/sandbox/deployment/test/ingress.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: sandbox-ui-ingress-test
```
stzanakis, line 4: can you add the prefix `metis-` so `metis-sandbox-ui-ingress-test`.  
Same goes for all namings.  
That will be more descriptive for us and others.

projects/sandbox/deployment/test/ingress.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
```
stzanakis, line 1: You can use the annotations and tls similarly to the ingress for metis ui.  
The metis-ui host/tls/secret `metis.test.eanadev.org` can in this case be perhaps `metis-sandbox.test.eanadev.org`.  
This applies to acceptance similarly as well.  
What do you think?

k8s-deploy.sh
```diff
@@ -95,6 +99,12 @@ then
   exit 1;
 fi
 
+# Check for invalid project
```
stzanakis, line 102: Just a comment. The other upcoming projects, e.g. metis-statistics-dashboard will also require some script for their deployment.
Something to think about on how perhaps this script could be reusable for you?

projects/sandbox/deployment/production/ingress.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
```
stzanakis, line 1: This will have to change as before, but we can handle that when we plan the migration for the production environment

projects/sandbox/deployment/production/sandbox-ui-service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
stzanakis, line 1: You don't have to name the file specifically `sandobx-ui-service.yml`. Like other files `ingress.yml`, `deployment.yml`, this could be `service.yml`.  
Or was that intended?  
Does not matter either way so it is up to you, just thought I should mention this.

k8s-deploy.sh
andyjmaclean: The stats dashboard is in another repository and so has its own script.  The projects share a deployment script here because it's a monorepo.

PR #736, opened 2023-03-02
MET-5123 Error Handling Refactor: add http interceptor

andyjmaclean:


projects/metis/src/app/dataset/depublication/depublication.component.ts
```diff
@@ -310,37 +312,30 @@ export class DepublicationComponent extends DataPollingComponent {
...
-  /*  @param {HttpErrorResponse} err - the fail response
-  */
-  onError(err: HttpErrorResponse): void {
-    this.errors.handleError(err);
+   * - set isSaving to false and initialise notification
+   *  @param {HttpErrorResponse} error
+   */
+  onError(error: HttpErrorResponse): void {
     this.isSaving = false;
-    console.error(`${err.status}: ${err.statusText}`);
```
JoanaCMS, line 321: Here as well I think can be usefull to keep the log

projects/metis/src/app/_services/error.interceptor.ts
```diff
@@ -0,0 +1,60 @@
...
+  /** shouldRetry
+   * Returns a predfined delay or throws the error passed
+   * @param { HttpErrorResponse } error
+   * @returns Observable<number>
+   **/
+  shouldRetry(error: HttpErrorResponse): Observable<number> {
+    const status = parseInt(`${error.status}`);
+    if (status === 401) {
+      this.expiredToken();
+    } else if (![200, 406].includes(status)) {
```
JoanaCMS, line 44: Why are we specifying the http status we are looking for?
Don't we want to generalise and check for 2xx and 4xx statuses?

projects/metis/src/app/login/login.component.ts
```diff
@@ -89,7 +89,7 @@ export class LoginComponent extends DataPollingComponent implements OnInit {
           },
           (err: HttpErrorResponse) => {
             this.notification = errorNotification(
-              err.status === 406
+              [406, 401, undefined].includes(err.status)
```
JoanaCMS, line 92: Again, don't we want to check for all 4xx statuses?

projects/metis/src/app/app.component.ts
```diff
@@ -113,13 +113,14 @@ export class AppComponent extends SubscriptionManager implements OnInit {
   */
   cancelWorkflow(): void {
     if (this.cancellationRequest) {
+      this.errorNotification = undefined;
       this.subs.push(
         this.workflows.cancelThisWorkflow(this.cancellationRequest.workflowExecutionId).subscribe(
           () => {
-            console.log('cancelling');
+            // successful cancellation request made
```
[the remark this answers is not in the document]
andyjmaclean: In general logs shouldn't be there unless debugging.  I think this statement was only there to stop the linter complaining about an empty function, but now it doesn't complain because of the comment.

projects/metis/src/app/app.component.ts
```diff
@@ -113,13 +113,14 @@ export class AppComponent extends SubscriptionManager implements OnInit {
...
           },
           (err: HttpErrorResponse) => {
-            this.errors.handleError(err);
+            this.errorNotification = httpErrorNotification(err);
```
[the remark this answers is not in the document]
andyjmaclean: The error is handled by showing a notification (a red message on the screen).  In general we don't want log messages in the code!

projects/metis/src/app/app.module.ts
```diff
@@ -138,6 +139,11 @@ import { SearchResultsComponent } from './search-results';
       provide: HTTP_INTERCEPTORS,
       useClass: TokenInterceptor,
       multi: true
+    },
+    {
+      provide: HTTP_INTERCEPTORS,
```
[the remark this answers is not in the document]
andyjmaclean: The "multi: true" is there to prevent that!

projects/metis/src/app/dataset/depublication/depublication.component.ts
andyjmaclean: It may be useful but it isn't best practice!

projects/metis/src/app/_services/error.interceptor.ts
andyjmaclean: 200 (success) is not the same as 204 (no content).  We care about what should behave differently, and I've kept the relevant codes the same as they were.

projects/metis/src/app/login/login.component.ts
andyjmaclean: The "not found" status is not the same as an unauthenticated status, so should be handled differently!

projects/metis/src/app/dataset/depublication/depublication.component.ts
JoanaCMS: It isn't best practice? Why do you say that?
Overall writing logs is a good practice since it gives a full view of the application's behavior which it helps tracing out errors/debugging when unexpected issues occur. For `info` level logs I can understand that we don't want to make the browser console noisy, but in this case for `error` type logs it is very useful. 

But from what you explained before, the method `httpErrorNotification` ensures that the error message is shown, so this is handled

JoanaCMS (approved):

PR #737, opened 2023-03-06
MET 5120 Modal Max Width

andyjmaclean:


shared-styles/assets/sass/pandora/components/_modal.scss
```diff
@@ -20,6 +20,7 @@ body.modal-open {
     background: $white;
     height: auto;
     left: 50%;
+    max-width: 90vw;
     position: absolute;
```
jeortizquan, line 24: does this property also helps?
`overflow-wrap: break-word;`

jeortizquan (approved):
with the viewport reduced the information is legible.
image

side note: check if the property mentioned also can help to improve the text that are displayed inside.

PR #740, opened 2023-03-28
Feat/met 5157 input redesign

andyjmaclean:


projects/sandbox/src/app/app.component.html
```diff
@@ -1,7 +1,12 @@
 <div class="header-wrapper">
   <header class="header">
     <div class="header-container">
-      <a class="hamburger" (click)="toggleSidebarOpen()"></a>
+      <a
+        class="hamburger"
```
JoanaCMS, line 5: What is `hamburger` supposed to be?

projects/shared/src/lib/form/file-upload/file-upload.component.html
```diff
@@ -1,8 +1,18 @@
-<input
-  #fileUpload
-  type="file"
-  accept="{{ acceptedTypes }}"
-  aria-label="File Upload"
-  [formGroup]="form"
-  [disabled]="form.disabled"
-/>
+<label
```
JoanaCMS, line 1: I don't know if this is the correct part of the code to point this out but I tested these changes locally and they look good, however I noticed that the upload file button did not react when I pressed the button, and because it took a few seconds to open the folder with the files, I was unsure if I had pressed the button, making me to press it again
Please make the button react when clicking on it.

projects/sandbox/src/app/app.component.html
JoanaCMS, line 5: What is hamburger supposed to be?

JoanaCMS (approved):

PR #743, opened 2023-04-05
MET-5170 Handle Dataset Errors

andyjmaclean:


projects/sandbox/src/app/_models/progress-info.ts
```diff
@@ -64,13 +64,19 @@ export interface TierInfo {
   total: number;
 }
 
+export interface DatasetLog {
+  type: string;
+  message: string;
+}
+
 export interface Dataset {
   status: DatasetStatus;
```
JoanaCMS, line 73: I see that `DatasetStatus` does not include the new value `FAILED`.
You should take it into account this new values. 
Perhaps put an error symbol next to the dataset name when it fails

projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
```diff
@@ -513,7 +513,16 @@ export class SandboxNavigatonComponent extends DataPollingComponent implements O
    *
    **/
   progressComplete(): boolean {
-    return !!this.progressData && this.progressData.status === DatasetStatus.COMPLETED;
+    const data = this.progressData;
+    if (data) {
+      if (data['dataset-logs'].length > 0) {
```
JoanaCMS, line 518: Check for the status instead, there  is a new status value from backend `FAILED` when the dataset processing fails

projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
```diff
@@ -608,7 +617,10 @@ export class SandboxNavigatonComponent extends DataPollingComponent implements O
 
         if (this.progressComplete()) {
           this.resetBusy();
-          if (this.progressData[fieldNamePortalPublish]) {
+          if (
+            this.progressData['dataset-logs'].length > 0 ||
```
JoanaCMS, line 621: Here you can also check with the status instead

JoanaCMS (changes requested):
I don't know where this is in the code but I did local tests and there's something I would like to be changed. When the dataset fails I see that there is an error symbol next to the date and time of the dataset, and to check the error I need to click the date and time instead of the error symbol, which I find counter intuitive.
I prefer to have the error symbol next to the dataset name instead which makes it consistent with how there is a tick when the dataset is finished processing
And to check the error message I prefer to click the title and/or the error symbol to check the message, I find it more intuitive this way

projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
andyjmaclean: Done

projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
andyjmaclean: Done

projects/sandbox/src/app/_models/progress-info.ts
andyjmaclean: Done

JoanaCMS (approved):

PR #756, opened 2023-05-03
MET-5250 Base tick / cross / warning on 'records-published-successfully'

andyjmaclean:


andyjmaclean:
I imagine reviewing this will go hand-in-hand with developing the back-end changes.

PR #759, opened 2023-05-04
MET-5261 Font / line-wrap tweaks

andyjmaclean:


projects/sandbox/src/app/problem-viewer/problem-viewer.component.html
```diff
@@ -350,7 +350,7 @@ <h2 class="title record-title">
           Record: {{ problemPatternsRecord.problemPatternList[0].recordAnalysisList[0].recordId }}
         </h2>
       </li>
-      <li>
+      <li *ngIf="!isLoading">
```
JoanaCMS, line 353: Is this option supposed to show up only when the dataset is fully published?
Because I still see it while the dataset is processing

andyjmaclean: This has been noticed by Deepti.  The check here is whether or not the problem data is ready, but that data is continually refreshed while the dataset is being processed.

To check whether the actual dataset processing is finished would require that we implement polling on the parent dataset, which gets quite complex in the case of problem patterns with records.

I decided to go with what we have, however, because it's no different to what we already show in the problem pattern viewer: that too needs to be refreshed, because it too can show stale data, because of the way that data processing and problem pattern processing run in parallel.

JoanaCMS (approved):

PR #769, opened 2023-05-31
MET-5309 Set correct 'busy' flags (Sandbox)

andyjmaclean:


projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
```diff
@@ -572,7 +559,7 @@ export class SandboxNavigatonComponent extends DataPollingComponent implements O
           this.problemPatternsDataset = undefined;
           confStep.error = err;
           confStep.lastLoadedIdDataset = undefined;
-          this.resetBusy();
+          confStep.isBusy = false;
```
jeortizquan, line 562: just out of curiosity `isProcessing` better word than `isBusy`?

jeortizquan (changes requested):
The feature is working. tested with a big dataset to replicate issue.
afbeelding
after checking in the tier calculation report the revolving cog is displayed
afbeelding

jeortizquan (approved):

projects/sandbox/src/app/sandbox-navigation/sandbox-navigation.component.ts
andyjmaclean: Well, it could be loading or polling or (coming soon) processing the result... in this case `isLoading` may be better, but `isBusy` is nice and generic!

PR #781, opened 2023-07-04
MET-5385 Add scroll hint / test data tweaks / grid line tweaks

andyjmaclean:


JoanaCMS (changes requested):
Overall the code looks good, however I do have some comments based on testing but I can't figure out where it is located in the code:
- The titles "Content Tier" and "Metadata Tier" in the table seem to have inconsistent structures. The title "Content Tier" seems to be towards to the left with an arrow to its right, while "Metadata Tier" title is all the way to the right with the arrow to its left. It looks a bit weird so I would like to see consistency. Perhaps if "Content Tier" follows the same structure as "Metadata Tier" it will help to remove the zig zag next to it.
- There's a behavior with the arrows that I find weird. For example, if I click on the table the word "Content Tier", it shows me the arrow, if I click in the arrow it reverses its direction, if I click again in the arrow it disappears. I was expecting for it to keep changing directions every time I click the arrow and for it to only disappear when I click "Content Tier" again. 
Please have a look into this

JoanaCMS (approved):

PR #809, opened 2023-10-24
MET-5564 Add Documentation Arrows

andyjmaclean:


projects/sandbox/src/app/app-routing.module.ts
```diff
@@ -31,6 +31,32 @@ const routes: Routes = [
       reuseComponent: true
     }
   },
+  {
+    path: 'x',
```
jeortizquan, line 35: sandbox easter egg :joy: 

https://elgoog.im/matrix/3d/

jeortizquan: afbeelding

projects/sandbox/src/app/doc-arrows/doc-arrows.component.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  arrowActiveKey(event: KeyboardEvent): void {
+    const tgt = event.target as HTMLElement;
+
+    if (tgt) {
+      const arrow = tgt.closest('.arrow') as HTMLElement;
+
+      if (arrow) {
+        // rotation
+        if (['r', 'R'].includes(event.key)) {
```
jeortizquan, line 97: only writing `r` works caps `R` at the moment is not possible to use.

afbeelding

jeortizquan (commented):
My suggestion would be to add a small tooltip, reminding user controls
```
use Arrows to change size.
SHIFT+R to rotate arrow.
CTRL+Arrow to move bigger step.
Backspace to delete arrow.
```

afbeelding

projects/sandbox/src/app/doc-arrows/doc-arrows.component.ts
andyjmaclean: If there's a diagram using letters up to 'R' then it probably has too many arrows!

But yes, they'd have to use cap lock in this case.

PR #824, opened 2024-01-11
MET-5689 Localised Creation Date

andyjmaclean:


projects/sandbox/src/app/dataset-info/dataset-info.component.ts
```diff
@@ -99,4 +105,14 @@ export class DatasetInfoComponent extends SubscriptionManager {
...
     this.subs.push(this.modalConfirms.open(this.modalIdProcessingErrors).subscribe());
   }
+
+  /**
+   * convertUTCDateToLocalDate
+   * @param {string} dateIn
+   **/
+  convertUTCDateToLocalDate(dateIn: string) {
+    const date = new Date(dateIn);
+    const newDate = new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000);
```
JoanaCMS, line 115: What are those magic numbers? 60 and 1000?
Could you create some class variables to represent what those numbers are?

projects/sandbox/src/app/dataset-info/dataset-info.component.ts
```diff
@@ -99,4 +105,14 @@ export class DatasetInfoComponent extends SubscriptionManager {
...
+    return newDate;
```
JoanaCMS, line 116: You can simply do `return new Date(date.getTime() - date.getTimezoneOffset() * 60 * 1000);`

projects/sandbox/src/app/dataset-info/dataset-info.component.ts
```diff
@@ -37,6 +37,12 @@ export class DatasetInfoComponent extends SubscriptionManager {
       this.sandbox
         .getDatasetInfo(datasetId, this.status !== DatasetStatus.COMPLETED)
         .subscribe((info: DatasetInfo) => {
+          let creationDate = info['creation-date'];
+          info['creation-date'] = formatDate(
+            this.convertUTCDateToLocalDate(creationDate),
+            'dd/MM/yyyy, HH:mm:ss',
+            'en-GB'
```
JoanaCMS, line 44: Is it correct to simply assume that it is always this locale?

projects/sandbox/src/app/dataset-info/dataset-info.component.ts
andyjmaclean: Local pre-computed variable added.

projects/sandbox/src/app/dataset-info/dataset-info.component.ts
andyjmaclean: The convert function assumes a non-localised / universal date (i.e. suffixed with something like the `+01:00` instead of the `Z`) as a parameter.  The 3rd parameter (`en-GB`) specifies the UI locale.

projects/sandbox/src/app/dataset-info/dataset-info.component.ts
andyjmaclean: Done.

JoanaCMS (approved):

PR #830, opened 2024-01-31
Angular 17 Update (Matomo Disabled)

andyjmaclean:


projects/metis/src/app/register/register.component.ts
```diff
@@ -15,11 +22,28 @@ import {
 } from '../_models';
 import { AuthenticationService, DocumentTitleService } from '../_services';
 import { TranslateService } from '../_translate';
+import { TranslatePipe } from '../_translate/translate.pipe';
```
jeortizquan, line 25: cleanup?

projects/sandbox/src/app/_models/navigation-orbs.ts
```diff
@@ -1,4 +1,5 @@
 export enum DisplayedTier {
+  NONE = -1,
   CONTENT = 0,
   METADATA
```
jeortizquan, line 4: My suggestion is that `METADATA` should also have an explicit value defined for readability.

jeortizquan (commented):
There are small observation, have a look

jeortizquan (approved):

PR #831, opened 2024-02-01
MET-5709 Handle Long Dataset Names

andyjmaclean:


jeortizquan (commented):
The buttons `View in Preview` and `View in Collections` are disconfigured after the update.

jeortizquan (approved):
The information is display accordingly.
afbeelding

PR #834, opened 2024-02-16
MET-5802 Remove Register User Feature

andyjmaclean:


jeortizquan (commented):
All change looks good I just saw some left overs in the translation file
`lang-en.ts` my suggestion is to remove them also
```
  register: 'Register to Metis',
  registrationAlready: 'You are already registered, you will be redirected to the sign in page!',
  registrationFailed: 'Registration failed, try again later',
  registrationSuccessful: 'Registration successful, you will be redirected to the sign in page!',
  userPasswordWeakError: 'Password is too weak',
```

Visually the feature is removed
afbeelding

jeortizquan (approved):

PR #867, opened 2024-11-12
MET-6236 Refactor Problem Pattern Scrolling

andyjmaclean:


projects/sandbox/src/app/problem-viewer/problem-viewer.component.ts
jeortizquan, line 1: now this can be renamed to `content-viewer/content-viewer.component.ts` because is used in two places now.

jeortizquan (approved):
A suggestion to rename the `problem-viewer` component to `content-viewer` because of the function that plays now on the application displaying different content for problem patterns and debias feature.

PR #888, opened 2025-02-26
MET-6357 Prevent User Tab From Crashing Layout

andyjmaclean:


jeortizquan (approved):
After reviewing the layout is fixed, the focus is lost because of the scrollbar.

PR #889, opened 2025-03-03
MET-6358 Refocus Dialog Openers

andyjmaclean:


projects/shared/src/lib/modal-confirm/modal-confirm.component.ts
```diff
@@ -66,15 +68,18 @@ export class ModalConfirmComponent implements ModalDialog, OnInit, OnDestroy {
       return;
     }
     if (e.key === 'Escape') {
+      this.keyboardClose = true;
       this.close(false);
     }
+    this.keyboardClose = false;
```
jochen-vermeulen, line 74: I don't think this line is needed? It will only be set to true just before calling the close function anyway.

In fact, `keyboardClose` could even be a parameter of the close() function? Instead of a class attribute?

PR #893, opened 2025-03-11
MET-6416-Sandbox-Accessibility-Improvements

andyjmaclean:


projects/sandbox/src/app/dataset-info/dataset-info.component.html
```diff
@@ -29,7 +29,7 @@
       }"
       [attr.title]="canRunDebias ? 'run debias report' : 'view debias report'"
       (click)="runOrShowDebiasReport(canRunDebias, openerDebias)"
-      (keydown.enter)="runOrShowDebiasReport(canRunDebias, openerDebias)"
+      (keydown.enter)="runOrShowDebiasReport(canRunDebias, openerDebias, true)"
```
jeortizquan, line 32: from my suggestion: boolean parameters are not a good practice. they are a code smell. usually means a function needs to be split into two different functions.

projects/sandbox/src/app/progress-tracker/progress-tracker.component.html
jeortizquan, line 1: I think one of the `a` elements don't need the focus.
afbeelding

jeortizquan: when going through the cycle you spot a orange line between the `step` and `view detail`

jeortizquan (commented):
Please have a look on the cycle on the progress tracker, and the suggestion of the boolean parameter.

projects/sandbox/src/app/progress-tracker/progress-tracker.component.html
andyjmaclean: This was coming from the fact that an entry in the model may contain errors that don't necessarily fail the step (i.e. warnings), and the template logic was hiding the link from the mouse, but not from the keyboard.

I chose to address this by using the :empty pseudo selector.  In theory the data could be analysed in the `.ts` and the model augmented, but I think this light-touch is more in keeping with scope as well as original approach.  I've added a explanitive comment in the [.scss](url).

jeortizquan (approved):
just the tab order when contentTier zero is enabled will work only backwards from the dataset number.

PR #895, opened 2025-03-23
Feat/met 6405 user accounts sandbox

andyjmaclean:


jeortizquan (changes requested):
There is a "bug" after login in the `upload dataset` icon is not displayed only the `create dataset is enabled`
afbeelding


afbeelding

jeortizquan (approved):
According to our conversation the icons behave as breadcrumb. The hard reload makes only the active icon to stay and the constants is fine for javascript.

PR #898, opened 2025-04-08
MET-5407 Prevent Logout on 406

andyjmaclean:


jeortizquan (approved):
a message is now displayed.

PR #907, opened 2025-05-15
MET-5783: Error 500 on Trasnformations - Use RxResurce

andyjmaclean:


jeortizquan (approved):
Just an observation. Is not displaying the error anymore.

PR #919, opened 2025-06-28
MET-6562 Add User Dataset Drop-In

andyjmaclean:


jeortizquan (approved):
suggestion, add blur filter background when showing the menu, so user knows where to focus.

afbeelding

PR #936, opened 2025-08-30
MET-6745 Sandbox Keycloak Session Handling

andyjmaclean:


projects/sandbox/src/app/debias/debias.component.spec.ts
```diff
@@ -106,24 +111,44 @@ describe('DebiasComponent', () => {
 
     it('should poll the debias report', fakeAsync(() => {
       expect(component.debiasReport).toBeFalsy();
-
-      component.datasetId = DebiasState.COMPLETED;
-
+      component.datasetId = '1'; //DebiasState.COMPLETED;
```
jeortizquan: suggesttion to cleanup the comment.

jeortizquan (changes requested):
The logout doesn't happen any more

but It doesn't update debias detection, it stays on zero, unless you click it again the real number appears.
see the examples below.

image

image

projects/sandbox/src/app/debias/debias.component.spec.ts
andyjmaclean: The above screenshots shows "run report" alongside the detection account, but these things should never be visible at the same time.

I think something has went wrong with either the code there or the set-up, Jorge!

I've tidied up the comment.

andyjmaclean: Done

jeortizquan (approved):

PR #945, opened 2025-10-13
MET-6763 Date-Time Display

andyjmaclean:


projects/sandbox/src/app/_data/static-data.ts
```diff
@@ -1,5 +1,8 @@
 import { DropInConfItem, ProblemPatternQualityDimension, ProblemPatternSeverity } from '../_models';
```
jeortizquan, line 2: my suggestion would be to uppercase this constants.
```
export const DATE_CONCISE_FMT = 'dd MMM (hh:mm a)';
export const DATE_VERBOSE_FMT = 'dd/MM/yyyy - hh:mm a';
```

projects/sandbox/src/app/drop-in/drop-in.component.ts
```diff
@@ -64,6 +64,8 @@ export class DropInComponent implements OnDestroy, OnInit {
   private readonly changeDetector = inject(ChangeDetectorRef);
   private readonly destroyRef = inject(DestroyRef);
 
+  public fmtDateConcise = fmtDateConcise;
```
jeortizquan: this is unused.

jeortizquan (changes requested):

projects/sandbox/src/app/_data/static-data.ts
andyjmaclean: Done.

projects/sandbox/src/app/drop-in/drop-in.component.ts
andyjmaclean: The service has already formatted the date, yes.  Deleted.

jeortizquan (approved):

PR #946, opened 2025-10-15
MET-6818 Sandbox Progress Display

andyjmaclean:


projects/sandbox/src/app/progress-tracker/progress-tracker.component.html
jeortizquan, line 1: observation it now displays the active step
image

projects/sandbox/test-data/index.ts
```diff
@@ -127,12 +127,14 @@ new (class extends TestDataServer {
...
 
+    const country = getParam('country');
+
     const data = this.initialiseGroupedDatasetData(
       `${this.newId}`,
       `${this.userId}`,
       harvestProtocol,
       datasetName,
-      getParam('country'),
+      country[0].toUpperCase() + country.slice(1).toLowerCase(),
```
jeortizquan, line 137: suggestion.
```
const capitalize = str =>
  typeof str === 'string' && str.length > 0
    ? str.length === 1
      ? str.toUpperCase()
      : str.charAt(0).toUpperCase() + str.slice(1).toLowerCase()
    : '';
const country = capitalize(getParam('country'));
```

jeortizquan (approved):
There is a suggestion see if that applies.

projects/sandbox/test-data/index.ts
andyjmaclean: Thanks - I may make a string utils function coded along these lines for use here and elsewhere.

PR #956, opened 2025-12-17
MET-6930 Handle 413 Upload Errors

andyjmaclean:


projects/sandbox/src/app/http-errors/errors.component.html
```diff
@@ -4,9 +4,13 @@
   <a class="close-errors" (click)="fnOnClose()"></a>
   }
   <div class="heading">
-    {{ obError.statusText }}
+    {{ obError.status }}
+    @if(obError.status === 404) { NOT FOUND } @if(obError.status === 413) { PAYLOAD TOO LARGE }
   </div>
   <ul class="error-detail">
+    @if(obError.status === 413){
+    <li>The upload exceded the file size limit of {{ uploadMaxSize }}</li>
```
jeortizquan: can you display the json message field? please. the backend fills that field for the UI.

projects/sandbox/src/app/http-errors/errors.component.ts
```diff
@@ -10,4 +10,5 @@ import { HttpErrorResponse } from '@angular/common/http';
 export class HttpErrorsComponent {
   readonly error = input<HttpErrorResponse>();
   readonly onClose = input<() => void>();
+  readonly uploadMaxSize = '64MB';
```
jeortizquan: my suggestion would be to display the message that comes from the backend. instead of hardcoding it on the UI.

jeortizquan (changes requested):
I would suggest to display `message` field from the json response.

jeortizquan (approved):

PR #959, opened 2025-12-30
MET-6977 Chained Rerun Design

andyjmaclean:


projects/sandbox/src/app/_services/sandbox-conf.service.ts
```diff
@@ -51,17 +53,23 @@ export class SandboxConfService {
   }
 
   isAncestorMode(): boolean {
-    return this.sandboxNavConf[2].stepSubClass === 'ancestor-mode';
+    return (this.sandboxNavConf[2].stepSubClass ?? '').includes(this.ANCESTOR_MODE);
```
jeortizquan, line 56: suggestion `this.sandboxNavConf[2]` can be a variable to be accessed with a name? e.g.
`this.sandboxNavConf[ITEM_NAME_BEING_ACCESSED]` where `const ITEM_NAME_BEING_ACCESSED=2`
or
`var navigationConfiguration = this.sandboxNavConf[2]`
saying this from readability perspective when coming back and read the code we know what that `2` item means.

jeortizquan (approved):
a  suggestion, see if this applies.

projects/sandbox/src/app/_services/sandbox-conf.service.ts
andyjmaclean: You're right.  I have plans to refactor this further in the future in the future.

